PR #59, opened 2015-04-20
updated to require region and region endpoint

EarthmanT:


aws-ec2/inputs.yaml.template
```diff
@@ -1,5 +1,7 @@
 aws_access_key_id: ''
 aws_secret_access_key: ''
+ec2_region_name: ''
+ec2_region_endpoint: ''
```
EarthmanT: bye bye endpoint

aws-ec2/inputs.yaml.template
```diff
@@ -1,5 +1,7 @@
 aws_access_key_id: ''
 aws_secret_access_key: ''
+ec2_region_name: ''
```
EarthmanT: not actually manditory

EarthmanT: check

aws-ec2/inputs.yaml.template
EarthmanT: check

PR #73, opened 2015-06-24
CFY-2918 add logrotate config

m1keil:


simple/components/frontend/scripts/create.sh
```diff
@@ -40,6 +40,26 @@ ctx logger info "Copying rest-location.cloudify file to /etc/nginx/conf.d/rest-l
...
+        daily
+        missingok
+        rotate 7
+        compress
+        delaycompress
+        notifempty
+        create 640 nginx adm
+        sharedscripts
+        postrotate
+                [ -f /var/run/nginx.pid ] && kill -USR1 `cat /var/run/nginx.pid`
```
nir0s: I guess USR1 is signaling nginx to rotate here, right?

m1keil: yep

PR #74, opened 2015-06-25
Cfy 2748 add more inputs to the openstack manager blueprint

isaac-s:


barakm:
The PR is outdated, as the new manager BP for RHEL/CentOS is very different from the current one.
I am going to close this PR, and we will re-visit the proposed changes when we the new openstack BP is released.

PR #86, opened 2015-07-01
CFY-2925 add logrotate to node

m1keil:


simple/components/webui/scripts/create.sh
```diff
@@ -46,4 +46,22 @@ ctx logger info "Deploying Grafana Configuration..."
 grafana_conf=$(ctx download-resource "components/webui/config/config.js")
 sudo mv ${grafana_conf} "${GRAFANA_HOME}/config.js"
 
+ctx logger info "Configuring logrotate..."
+lconf="/etc/logrotate.d/nodejs"
```
nir0s: This is quite like doing "/etc/logrotate.d/python". These are not nodejs logs but rather the webui's.

simple/components/webui/scripts/create.sh
```diff
@@ -46,4 +46,22 @@ ctx logger info "Deploying Grafana Configuration..."
...
+
+cat << EOF | sudo tee $lconf > /dev/null
+$WEBUI_LOG_PATH/*.log {
```
nir0s: I actually think that I saw that the log4js config in gsPresets.json already rotates logs. Are there any additional logs to rotate?

PR #97, opened 2015-08-11
CFY-3012 reconfigure daemons running under root

m1keil:
daemons running under root user were reconfigured to run with own dedicated user

### DIFF
--- a/simple/components/restservice/config/cloudify-restservice.service
+++ b/simple/components/restservice/config/cloudify-restservice.service
@@ -10,6 +10,8 @@ ExecStart=/bin/sh -c '/opt/manager/env/bin/gunicorn \
     --pid /var/run/gunicorn.pid \
     -w $(($(nproc)*2+1)) \
     -b 0.0.0.0:${REST_SERVICE_PORT} \
+    --user restservice \
+    --group restservice \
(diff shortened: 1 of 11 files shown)


simple/components/restservice/config/cloudify-restservice.service
```diff
@@ -6,8 +6,10 @@ PIDFile=/var/run/gunicorn.pid
 TimeoutStartSec=0
 Restart=on-failure
 EnvironmentFile=-/etc/sysconfig/cloudify-restservice
+User=rest
```
nir0s: can you call this user restservice to correspond with the naming convention?

simple/components/restservice/config/cloudify-restservice.service
```diff
@@ -6,8 +6,10 @@ PIDFile=/var/run/gunicorn.pid
...
+Group=rest
 ExecStart=/bin/sh -c '/opt/manager/env/bin/gunicorn \
-    --pid /var/run/gunicorn.pid \
+    --pid /var/run/gunicorn/gunicorn.pid \
```
nir0s: odds are slim.. but we need to make sure this file isn't being read anywhere from within our code.

nir0s:
LGTM

PR #101, opened 2015-08-18
CFY-3012 reconfigure daemons running under root

nir0s:
daemons running under root user were reconfigured to run with own dedicated user


nir0s:
Please solve conflicts.

simple/components/mgmtworker/config/cloudify-mgmtworker.service
```diff
@@ -5,6 +5,8 @@ Description=Cloudify Management Worker Service
 TimeoutStartSec=0
 Restart=on-failure
 EnvironmentFile=-/etc/sysconfig/cloudify-mgmtworker
+User=celery
```
nir0s: I'm wondering about this.
When we bootstrap, we upload a pem file for the agents to use. When we upload it, we upload it to the bootstrapping user's home directory. Since celery is now the user for the management worker which later uses that pem file to access to host machines, we need to verify that it can access the uploaded pem file.

PR #105, opened 2015-09-03
Cfy 3066 enable idempotency and add stop lifecycle operation

nir0s:


new/simple-manager-blueprint.yaml
```diff
@@ -81,7 +81,7 @@ inputs:
   #############################
   java_source_url:
     type: string
-    default: https://dl.dropboxusercontent.com/u/407576/3.2/java.rpm
+    default: https://dl.dropboxusercontent.com/u/407576/3.3/jre1.8.0_45-1.8.0_45-fcs.x86_64.rpm
```
m1keil: dropbox?

new/components/utils
```diff
@@ -169,16 +167,56 @@ function yum_install
...
+    # separated array of packages as you can with yum install. You must
+    # provide one package per invocation.
     ###
     source=$1
+    source_name=$(get_file_name $source)
+    extension="${source_name##*.}"
+    archive_path=$source_name
+
+    if [ "$extension" == "rpm" ]; then
+        archive_path="/etc/cloudify/$source_name"
```
m1keil: why in /etc ???

new/components/utils
```diff
@@ -128,6 +107,21 @@ function download_file
     echo ${destination_file}
 }
 
+function download_cloudify_resource() {
+    url=$1
+    destination_file="/etc/cloudify/$(get_file_name $url)"
```
m1keil: add variable DOWNLOAD_PREFIX or something similar instead of hardcoding this, will be easier to change

new/simple-manager-blueprint.yaml
nir0s: this is changed with the https PR.

new/components/utils
nir0s: changed to /opt

PR #112, opened 2015-09-09
Cfy 3221 use restservice code rpm

nir0s:


new/simple-manager-blueprint.yaml
```diff
@@ -36,52 +36,68 @@ inputs:
...
+    type: string
+    default: http://gigaspaces-repository-eu.s3.amazonaws.com/org/cloudify3/3.3.0/m5-RELEASE/cloudify-management-worker-3.3.0-m5_b275.x86_64.rpm
+
+  amqpinflux_rpm_source_url:
+    type: string
+    default: http://gigaspaces-repository-eu.s3.amazonaws.com/org/cloudify3/3.3.0/m5-RELEASE/cloudify-amqp-influx-3.3.0-m5_b275.x86_64.rpm
+
+  cloudify_resources_url:
+    type: string
+    default: https://github.com/cloudify-cosmo/cloudify-manager/archive/3.3m5.tar.gz
```
m1keil: shouldn't this and all the above be master branch?

new/openstack-manager-blueprint.yaml
```diff
@@ -117,52 +117,68 @@ inputs:
...
-    default: https://github.com/cloudify-cosmo/cloudify-diamond-plugin/archive/master.tar.gz
+    default: ''
 
   #############################
   # Java Inputs
   #############################
   java_source_url:
     type: string
-    default: https://dl.dropboxusercontent.com/u/407576/3.2/java.rpm
+    default: https://dl.dropboxusercontent.com/u/407576/3.3/jre1.8.0_45-1.8.0_45-fcs.x86_64.rpm
```
m1keil: this needs to be in s3 right?

new/simple-manager-blueprint.yaml
nir0s: the cloudify_resources_url, yes.
Ths above are rpm's specific to the version.. like any other package (agent, cli, etc..), so no.

new/components/utils
```diff
@@ -39,10 +39,10 @@ function install_module
 
     if [[ ! -z "${venv}" ]]; then
         ctx logger info "Installing ${module} in virtualenv ${venv}..."
-        sudo ${venv}/bin/pip install ${module} >/dev/null
+        sudo ${venv}/bin/pip install ${module} --upgrade >/dev/null
```
m1keil: you sure about this one? maybe `--ignore-installed` instead?

nir0s: I don't want it to ignore installed. If a developer (or whoever) provided a branch to a module, it means they want to use that branch instead of the already installed one. This allows to upgrade that module.

PR #127, opened 2015-09-22
CFY-3629 Added support to installing plugins on rest service

max-orlov:


new/components/restservice/scripts/install_plugins.py
```diff
@@ -0,0 +1,78 @@
...
+import fabric
+import fabric.api
+import jinja2
+
+from cloudify import ctx
+
+
+def install_plugins():
+
+    root_dir = os.path.join(ctx._endpoint.storage.resources_root,
```
dankilman: no need for `ctx._endpoint.storage.resources_root`. When you `os.path.join`, and the n'th component is an absolute path (such as the case with `os.path.dirname(__file__)` the previous parts are imply discarded. So you can just do
`root_dir = os.path.dirname(__file__)`

new/components/restservice/scripts/install_plugins.py
```diff
@@ -0,0 +1,78 @@
...
+    ctx.logger.info('Installing plugins')
+
+    # Getting all the required plugins
+    plugins = ctx.node.properties.get('plugins', {})
+
+    # Shame to do all the work for nothing
+    if plugins:
+
+        # # Get components utils (for pip install)
+        with open(os.path.normpath(
```
dankilman: you can simply do `utils_content = ctx.get_resource('components/utils')` instead of all this

new/components/restservice/scripts/install_plugins.py
```diff
@@ -0,0 +1,78 @@
...
+                with tarfile.open(fileobj=fileobj, mode='w:gz') as tar:
+                    tar.add(plugin_path, arcname=name)
+                fileobj.seek(0)
+                plugin['source'] = \
+                    'file://{0}/{1}.tar.gz'.format(cloudify_plugins, name)
+                fabric.api.put(local_path=fileobj,
+                               remote_path=plugin['source'])
+
+        # render script template and copy it to host's home dir
+        with open('{0}/{1}'.format(root_dir, install_plugins_script)) as\
```
dankilman: again, you can simply do `script_template = ctx.get_resource('components/restservice/scripts/install_plugins.sh')`

new/components/restservice/scripts/install_plugins.py
```diff
@@ -0,0 +1,78 @@
...
+                template_file:
+            script_template = template_file.read()
+        script = jinja2.Template(script_template).render(plugins=plugins)
+        fabric.api.put(local_path=StringIO(script),
+                       remote_path='~/{0}'.format(install_plugins_script))
+
+        _run_command('sudo chmod +x ~/{0}'.format(install_plugins_script))
+
+        # Execute the rendered script
+        _run_command('~+/{0}'.format(install_plugins_script))
```
dankilman: what does the `+` stand for?

PR #128, opened 2015-09-22
CFY-3629-Added support for installing plugins on rest service in the new manager blueprint.

max-orlov:


new/components/restservice/scripts/install_plugins.py
```diff
@@ -0,0 +1,64 @@
...
+        # render script template and copy it to host's home dir
+        script_template = ctx.get_resource('components/restservice/'
+                                           'scripts/install_plugins.sh')
+        script = jinja2.Template(script_template).render(plugins=plugins)
+        fabric.api.put(local_path=StringIO(script),
+                       remote_path='~/cloudify/{0}'
+                       .format(install_plugins_script))
+
+        # Execute the rendered script
+        _run_command('sudo chmod +x ~/cloudify/{0} && ./cloudify/{0}'
```
dankilman: I think it better being explicit about the script path, it is also consistent with the path being `chmod`ed
`sudo chmod +x ~/cloudify/{0} && ~/cloudify/{0}` (instead of `./cloudify`)
Also, no need for sudo here

PR #131, opened 2015-09-29
CFY-3224-Support-rabbitmq-security

geokala:


new/components/amqpinflux/scripts/create.sh
```diff
@@ -27,4 +44,5 @@ sudo useradd --shell /sbin/nologin --home-dir "${AMQPINFLUX_HOME}" --no-create-h
 ctx logger info "Fixing permissions..."
 sudo chown -R "${AMQPINFLUX_USER}:${AMQPINFLUX_GROUP}" "${AMQPINFLUX_HOME}"
 
-configure_systemd_service "amqpinflux"
\ No newline at end of file
+configure_systemd_service "amqpinflux"
+inject_service_env_var "(( amqp_cert_path ))" "${AMQP_CERT_PATH}" "amqpinflux"
```
nir0s: If would try and refrain from using inject_service_env_var as it's kinda hackish and set a runtime property and use it using download-resource-and-render. something like this:

```shell
# setting the property
ctx instance runtime_properties amqp_cert_path "${AMQP_CERT_PATH}"
```

and then inside the new/components/amqpinflux/config/cloudify-amqpinflux file write:

AMQP_CA_CERT_PATH="{{ ctx.instance.runtime_properties.amqp_cert_path }}"

This would be cleaner in the long run.

new/types/manager-types.yaml
```diff
@@ -561,6 +617,18 @@ node_types:
...
         default: { get_input: amqpinflux_source_url }
+      rabbitmq_username:
+        description: RabbitMQ username for access to queues.
+        type: string
+        default: { get_input: rabbitmq_username }
+      rabbitmq_password:
+        description: RabbitMQ password for access to queues.
+        type: string
+        default: { get_input: rabbitmq_password }
+      rabbitmq_ca_cert:
```
nir0s: I see we have different names for this property in different node types. maybe they should be the same? easier to find and handle...

new/components/riemann/scripts/create.sh
```diff
@@ -69,3 +68,5 @@ deploy_blueprint_resource "${CONFIG_REL_PATH}/main.clj" "${RIEMANN_CONFIG_PATH}/
 # we inject the management_ip for both of these to Riemann's systemd config. These should be potentially different
 # if the manager and rabbitmq are running on different hosts.
 configure_systemd_service "riemann"
+inject_service_env_var "{{ ctx.node.properties.rabbitmq_username }}" "$(ctx node properties rabbitmq_username)" "riemann"
```
nir0s: use download-resource-and-render for these

new/components/restservice/config/cloudify-rest.conf
```diff
@@ -6,10 +6,13 @@
     rest_service_log_files_backup_count: 20,
     db_address: '{{ ctx.node.properties.es_endpoint_ip }}',
     db_port: '{{ ctx.node.properties.es_endpoint_port }}',
-    amqp_address: 'localhost',
+    amqp_address: 'localhost:(( amqp_port ))/',
```
nir0s: again, I think we should use runtime properties and download-resource-and-render for these instead of injecting them.

new/components/amqpinflux/scripts/create.sh
noakup: :+1:

new/components/manager/config/provider_context
```diff
@@ -1 +1 @@
...
...dmC1iIN2o7r6wO8tdr+k1Q1rUjQpv8U8+XlEQvIjSBMYNVLSyo47Wx\\nYvLZQpB02UL4kJFFAnM8drqqNpP+h2dqsoLJwBxfaajmv9nKIdiCRIK5CEaFc2uta1K6MWX8tmRu\\n3gHTUddS0+E+dKmmQt2XNJWUv3UC7NNhPR3uvl3W8gy/sNO1vyTFX5LiL0nxycj8JSk+e5a/JMVf\\nkuIvSfGXpPhLUvwlKf6SFH9Jir8kxV+S4i9J2XpJysku5/BXrBzxipX2ev4+FJ+Cwqeg8PeheHHx\\n4uLFxd+H4u9D8feh+PtQ/H0o/j4Ufx+Kt8b8fSj+PhRvsvn7UPx9KP4SA3+Jgb/EwN+H4lWJVyVe\\nlfj7UHy2IH8fir8Pxd+H4u9DOWoOJO8a9K5Bf7XG+Vyt4a8kehdXEtXPstfSydnsEV0Z5ezj4ph1\\nLUfJG+RYcQANmgAVqVbqJ5DL0/GNk8mV09+bR49/czkfdC1dSb0152IjH0hH4oxPJlWGzonhGptR\\nZKEOdV4b9LSE/0jC4S2BCmXxJXHZNCRAXFZnEhUZIhBL1NHpBRBJ2hwajewx7fiooamyplg4TIYS\\nNXtV9UUVvdGwLWvGJ10F2SoGj5g8qDwwgv4vB86pTixr0ccPnz4rgGNDRXPKPchlxuNutmyna1kN\\nAfmQzpEAVGVJneLla18SGSxpTBQTSBDQFJhgCacM0Rdh/2thuR5CAV0ufHHFKYCUrCJOwqP0AWwc\\n3t7eqM/RdDKqfppyO74YjW9up8PJaDwaXwxHk9H05gINT0GAXM3XCF3Yza/O93Y9t6gUn++kPP+A\\nyqmtmWdjM73KD+sfvMx8i/KvWH9VGJODry7/0+mkW/5BXJz8K10wHN1OrqcXaOLl/634X5P6I+v/\\n0fXwtq7/x8Pbm6nX/yfR/4dkBXuZ53FXEjG9IGpxHNqlbjMirZaHrR4QZNMZidpy6lRegiIgqSUU\\nqXjQT2n85v4Ea3abeFjp8mCafHy59uc7Grckz3ePtgRcVbcJdOMqlqkzsVuZVW26T3a41qRvfh4/\\nkv4vaX9E+/96OKrM//DeeDwcT7z+fyP9f7/lBG+7Tt6UeLMDuCWD5aukvjp5eBH2xRdffPHFF198\\n8cUXX3zxxRdffNla/g/Hv4N/AMgAAA==\\n"}
```
noakup: regardless of this commit, I'm not sure I understand why we need this... let's discuss :)

new/components/manager/scripts/configure_manager.sh
```diff
@@ -7,10 +7,14 @@ function _post_provider_context() {
...
 
+    sed_delimiter=$'\001'
+    cloudify_agent="$(ctx --json-output node properties cloudify.cloudify_agent)"
+    export PROVIDER_CONTEXT_DATA=$(echo "${PROVIDER_CONTEXT_DATA}" | sed s"${sed_delimiter}{{ cloudify_agent }}${sed_delimiter}${cloudify_agent//\\/\\\\}${sed_delimiter}")
+
     ctx instance runtime_properties manager_provider_context "$PROVIDER_CONTEXT_DATA"
 
     ctx logger info "Posting Provider Context..."
-    curl --fail --silent --request POST --data @${PROVIDER_CONTEXT_FILE} http://localhost/api/v2/provider/context --header "Content-Type:application/json" >/dev/null
+    curl --fail --silent --request POST --data "${PROVIDER_CONTEXT_DATA}" http://localhost/api/v2/provider/context --header "Content-Type:application/json" >/dev/null
```
noakup: I think we should add "--show-error" here and wherever we use curl --silent, so we will at least know why we fail, if we fail

new/components/mgmtworker/scripts/create.sh
```diff
@@ -52,4 +81,12 @@ if [ ! -z ${REST_SERVICE_SOURCE_URL} ]; then
     install_module "/tmp/workflows" ${MGMTWORKER_VIRTUALENV_DIR}
 fi
 
+extract_github_archive_to_tmp ${manager_repo}
+ctx logger info "Configuring Management worker..."
+deploy_blueprint_resource "${CONFIG_REL_PATH}/worker_conf.py" "${CELERY_WORK_DIR}/worker_conf.py"
+replace "(( broker_port ))" "${BROKER_PORT}" "${CELERY_WORK_DIR}/worker_conf.py"
```
noakup: don't we have a download_and_render that does the replacing?

new/components/rabbitmq/config/cloudify-rabbitmq.service
```diff
@@ -6,7 +6,8 @@ TimeoutStartSec=0
 Restart=on-failure
 EnvironmentFile=-/etc/sysconfig/cloudify-rabbitmq
 ExecStart=/usr/sbin/rabbitmq-server
+ExecStop=/usr/local/bin/kill-rabbit
```
noakup: oh so cruel! :rabbit:

new/components/rabbitmq/scripts/create.sh
```diff
@@ -45,26 +50,96 @@ EOF
...
+    ctx logger warn "Public certificate did not appear to be in PEM format."
+  fi
+  if [[ -f /etc/rabbitmq/rabbit-priv.pem ]]; then
+    ctx logger error "Private certificate was provided but public certificate was invalid."
+  fi
+fi
+
+# Configure rabbit
+if [[ -f /etc/rabbitmq/rabbit-priv.pem ]] && [[ -f /etc/rabbitmq/rabbit-pub.pem ]]; then
+  echo "[
```
noakup: well... there must be an easier way...

new/components/rabbitmq/scripts/create.sh
```diff
@@ -45,26 +50,96 @@ EOF
 
 sudo chmod 644 $lconf
 
+# Creating rabbitmq systemd stop script
+cat << EOF | sudo tee /usr/local/bin/kill-rabbit > /dev/null
```
noakup: it's duck season

new/components/manager/config/provider_context
geokala: I think this is actually from the mods I made following @nir0s 's initial new manager deployment which curled this up to the manager. I believe it is actually fully bypassed now, but would need to check with @nir0s

new/components/manager/scripts/configure_manager.sh
geokala: I think this is actually from the mods I made following @nir0s 's initial new manager deployment which curled this up to the manager. I believe it is actually fully bypassed now, but would need to check with @nir0s 

The function this bit of code is in is not run any more, from memory.

new/components/mgmtworker/scripts/create.sh
geokala: We do, it was just added after I did the initial mods on this. I've noted that I need to go through and runtime-attribute these so that they can be rendered in the same way as the rest.

new/components/rabbitmq/config/cloudify-rabbitmq.service
geokala: They're vicious creatures. It's self defence.

new/components/rabbitmq/scripts/create.sh
geokala: It probably wants commenting, and likely the certificate deployment wants turning into functions so that logic to check that they're actually being created as PEM formatted SSL certs exists. Otherwise, the only easier way would actually be to remove any checks I think?

new/components/rabbitmq/scripts/create.sh
geokala: It's just about to be. Starts in October.

new/components/logstash/scripts/create.sh
```diff
@@ -22,6 +24,8 @@ yum_install ${LOGSTASH_SOURCE_URL}
 
 ctx logger info "Deploying Logstash conf..."
 deploy_blueprint_resource "${CONFIG_REL_PATH}/logstash.conf" "${LOGSTASH_CONF_PATH}/logstash.conf"
+replace "{{ ctx.node.properties.rabbitmq_username }}" "${RABBITMQ_USERNAME}" "${LOGSTASH_CONF_PATH}/logstash.conf"
```
nir0s: download-resource-and-render

new/components/manager/config/provider_context
nir0s: this file should be deleted. we set the provider context within the blueprint or from the cli (depending on the key).

new/simple-manager-blueprint.yaml
```diff
@@ -331,6 +347,9 @@ node_templates:
           max_workers: 5
           remote_execution_port: 22
           user: { get_input: agents_user }
+          broker_user: { get_input: rabbitmq_username }
+          broker_pass: { get_input: rabbitmq_password }
+          broker_ssl_cert: { get_input: rabbitmq_cert_public }
```
nir0s: you also need to do the same for the openstack blueprint and add the SSL enabled input.

new/components/mgmtworker/config/cloudify-mgmtworker.service
```diff
@@ -5,20 +5,22 @@ Description=Cloudify Management Worker Service
 TimeoutStartSec=0
 Restart=on-failure
 EnvironmentFile=-/etc/sysconfig/cloudify-mgmtworker
+WorkingDirectory=/opt/mgmtworker/work
 ExecStart=/opt/mgmtworker/env/bin/celery worker \
     -Ofair \
     --include=cloudify_system_workflows.deployment_environment,cloudify_agent.operations,cloudify_agent.installer.operations,riemann_controller.tasks,cloudify.plugins.workflows \
     --broker=${BROKER_URL} \
```
nir0s: this should probably be removed as it is set within the worker_conf.py file.

new/components/mgmtworker/scripts/create.sh
```diff
@@ -19,6 +21,12 @@ export MGMTWORKER_VIRTUALENV_DIR="${MGMTWORKER_HOME}/env"
 export CELERY_WORK_DIR="${MGMTWORKER_HOME}/work"
 export CELERY_LOG_DIR="/var/log/cloudify/mgmtworker"
 
+export RABBITMQ_USERNAME="$(ctx node properties rabbitmq_username)"
+export RABBITMQ_PASSWORD="$(ctx node properties rabbitmq_password)"
+export RABBITMQ_CERT_PUBLIC="$(ctx node properties rabbitmq_cert_public)"
+
+# Currently assuming local host, not distributed
+export MANAGEMENT_IP=$(ctx instance host_ip)
```
nir0s: this can probably be removed.. please check.

new/components/mgmtworker/scripts/create.sh
```diff
@@ -40,11 +48,32 @@ ctx logger info "Installing Optional Management Worker Modules..."
...
+else
+  BROKER_CERT_PATH=""
+  if [[ -z "${RABBITMQ_CERT_PUBLIC}" ]]; then
+    ctx logger info "No public certificate found. TLS not enabled."
+  else
+    ctx logger warn "Public certificate did not appear to be in PEM format. TLS not enabled."
+  fi
+fi
+
+if [[ -z "${BROKER_CERT_PATH}" ]]; then
```
nir0s: this should also be deduced from whether SSL is enabled or not.

new/components/mgmtworker/scripts/create.sh
```diff
@@ -52,4 +81,12 @@ if [ ! -z ${REST_SERVICE_SOURCE_URL} ]; then
     install_module "/tmp/workflows" ${MGMTWORKER_VIRTUALENV_DIR}
 fi
 
+extract_github_archive_to_tmp ${manager_repo}
+ctx logger info "Configuring Management worker..."
+deploy_blueprint_resource "${CONFIG_REL_PATH}/worker_conf.py" "${CELERY_WORK_DIR}/worker_conf.py"
```
nir0s: we can remove the worker_conf.py file entirely and just deal with generating the yaml properties file for celery.

new/components/nginx/agent-files/celeryd-cloudify.conf.template
```diff
@@ -6,7 +6,7 @@ WORKER_MODIFIER="{{ worker_modifier }}"
 
 export BROKER_IP="{{ broker_ip }}"
 export MANAGEMENT_IP="{{ management_ip }}"
-export BROKER_URL="amqp://guest:guest@${BROKER_IP}:5672//"
+export BROKER_URL="amqp://{{ ctx.nodes.properties.rabbitmq_username }}:{{ ctx.nodes.properties.rabbitmq_password }}@${BROKER_IP}:5672//"
```
nir0s: this file is irrelevant and should be removed as it is a part of cloudify-agent itself.

new/components/rabbitmq/scripts/create.sh
```diff
@@ -45,26 +50,96 @@ EOF
...
+#! /usr/bin/env bash
```
nir0s: can we move this to a file and download_cloudify_blueprint_resource it..

new/components/rabbitmq/scripts/create.sh
```diff
@@ -45,26 +50,96 @@ EOF
...
+  fi
+  if [[ -f /etc/rabbitmq/rabbit-priv.pem ]]; then
+    ctx logger error "Private certificate was provided but public certificate was invalid."
+  fi
+fi
+
+# Configure rabbit
+if [[ -f /etc/rabbitmq/rabbit-priv.pem ]] && [[ -f /etc/rabbitmq/rabbit-pub.pem ]]; then
+  echo "[
+ {ssl, [{versions, ['tlsv1.2', 'tlsv1.1']}]},
```
nir0s: please do not generate, copy :)

new/components/restservice/config/cloudify-rest.conf
```diff
@@ -6,10 +6,13 @@
...
+    amqp_username: '{{ ctx.node.properties.rabbitmq_username }}',
+    amqp_password: '{{ ctx.node.properties.rabbitmq_password }}',
+    amqp_ca_path: '(( amqp_ca_cert_path ))',
```
nir0s: same thing

new/components/restservice/scripts/create.sh
```diff
@@ -33,6 +37,26 @@ create_dir ${REST_SERVICE_HOME}
 create_dir ${REST_SERVICE_LOG_PATH}
 create_dir ${MANAGER_RESOURCES_HOME}
 
+# Create certs as required
```
nir0s: we do this in multiple places. please write a function for it.

new/components/mgmtworker/scripts/create.sh
```diff
@@ -40,11 +47,23 @@ ctx logger info "Installing Optional Management Worker Modules..."
 [ -z ${SCRIPT_PLUGIN_SOURCE_URL} ] || install_module ${SCRIPT_PLUGIN_SOURCE_URL} ${MGMTWORKER_VIRTUALENV_DIR}
 [ -z ${AGENT_SOURCE_URL} ] || install_module ${AGENT_SOURCE_URL} ${MGMTWORKER_VIRTUALENV_DIR}
 
+# Add certificate and select port, as applicable
+if [[ -n "${RABBITMQ_CERT_PUBLIC}" ]]; then
```
geokala: This block of code is performed several times, but I'm uncertain that it should be farmed out to a function as it sets two runtime properties, which feels like the sort of side effect a utils function shouldn't have... could go either way on this though.

dankilman:
what's up with travis?

new/components/amqpinflux/scripts/create.sh
```diff
@@ -26,7 +28,18 @@ yum_install ${AMQPINFLUX_RPM_SOURCE_URL}
 
 create_service_user ${AMQPINFLUX_USER} ${AMQPINFLUX_HOME}
 
+if [[ "$(ctx node properties rabbitmq_ssl_enabled)" == 'True' ]]; then
```
nir0s: as per our convention, please assign this to a global variable.
Also, you can instead do `ctx -j` to get the boolean value.

new/components/logstash/scripts/create.sh
```diff
@@ -14,6 +14,12 @@ ctx instance runtime_properties es_endpoint_ip ${ES_ENDPOINT_IP}
 export LOGSTASH_LOG_PATH="/var/log/cloudify/logstash"
 export LOGSTASH_CONF_PATH="/etc/logstash/conf.d"
 
+# Confirm username and password have been supplied for broker before continuing
+# Components other than logstash and riemann have this handled in code already
+if [[ -z "$(ctx node properties rabbitmq_username)" ]] ||
```
nir0s: please assign these to global variables.

nir0s:
I just realized that you hadn't added the same configuration to the aws and openstack manager blueprints.. can you please add the relevant inputs and properties there?

geokala:
Hmm... thought I did that- I'll double check.

PR #134, opened 2015-09-30
CFY-3038 Manager import snapshot REST API

xwombat:


new/components/nginx/config/default.conf
```diff
@@ -1,5 +1,5 @@
 # Declares the size of a request's body. This is mainly used to allow large blueprints to be uploaded.
-client_max_body_size 50m;
+client_max_body_size 0m;
```
nir0s: Isn't this rather dangerous? If by mistake someone tries to download or upload a 20GB file.. that might create some problems. Maybe we can agree on a logical (albeit arbitrary) limit?

new/components/nginx/config/rest-location.cloudify
```diff
@@ -1,4 +1,4 @@
-location ~ ^/api/v2/(blueprints|executions|deployments|nodes|events|search|status|provider|api|node-instances|version|evaluate|deployment-modifications|tokens) {
+location ~ ^/api/v2/(blueprints|snapshots|executions|deployments|nodes|events|search|status|provider|api|node-instances|version|evaluate|deployment-modifications|tokens) {
```
nir0s: @m1keil, @idanmo, as a general comment, I think we should discuss this, but snapshot is a misnomer. We're not taking snapshots, but rather a full backup. A snapshot is a relative point in time. When we do provide incremental backups (and then, we'll probably call them snapshots), this will bite us in the ass API-wise.

m1keil: agreed. it's a bad and misleading term. we are not snapshotting, we are creating backup.

new/components/nginx/config/default.conf
simon-bar: A limit is a good approach. The thing is that metrics tend to be really heavy. We can change it in some future PR is needed.

PR #135, opened 2015-10-01
SCOR-216: vcloud manager blueprint

0lvin:


new/components/vcloud/update-vcloud-vm.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+
+uname -a
+
+# update system
+yum update --skip-broken -y
```
nir0s: why are these required?
We want to provide an 'offline' experience meaning that yum update might not be accessible. Also, installing compilers is in general something that we're trying to avoid, thanks to which we're providing all compilation-requiring python modules already compiled as wheels for Centos 7. Is this because the vcloud plugin itself required compilation? If so, @idanmo , please advise.

0lvin: About compilers: vcloud plugin required complier for build several modules that don't exist in ubuntu repositories,  so when we installed old manager and tried to use with vcloud we had issues with it. 

This blueprint have not fully tested with vcloud plugin deployments, so i will ask separately about merge later after full test and maybe we will drop requirements to install compilers/update system.

0lvin: gcc - required for lxml(https://github.com/vmware/pyvcloud/blob/master/requirements.txt#L2) package.

new/components/vcloud/configure.py
```diff
@@ -0,0 +1,72 @@
...
+        copy configuration to managment host,
+        install docker
+        and save current context to .cloudify/context
+        For now - we have saved only managment network name
+    """
+    _copy_vsphere_configuration_to_manager(vcloud_config)
+    _save_context()
+
+
+def _copy_vsphere_configuration_to_manager(vcloud_config):
```
kostya13: Change function name to:
_copy_vcloud_configuration_to_manager

new/vcloud-manager-blueprint.yaml
```diff
@@ -0,0 +1,867 @@
+tosca_definitions_version: cloudify_dsl_1_2
+
+imports:
+  - http://www.getcloudify.org/spec/cloudify/3.3m7/types.yaml
+  - http://www.getcloudify.org/spec/fabric-plugin/1.3m7/plugin.yaml
+  - types/manager-types.yaml
+  - https://raw.githubusercontent.com/cloudify-cosmo/tosca-vcloud-plugin/1.3m6/plugin.yaml
```
ran-z: why is this one m6?

new/components/vcloud/configure.py
```diff
@@ -0,0 +1,72 @@
...
+import fabric
+
+import vcloud_plugin_common
+from cloudify import ctx
+
+PROVIDER_CONTEXT_RUNTIME_PROPERTY = 'provider_context'
+
+
+def configure(vcloud_config):
+    """
```
nir0s: just to follow up on the standard, docs string should be written as:

```python
def func():
    """one liner summary

    additional info
    """
```

not a must. just sayin`..

new/components/vcloud/update-vcloud-vm.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -e
```
nir0s: please move this script to the "manager" folder. We're trying to keep things as clean as possible. Anything that is related to the manager as an entity rather to one of the components should go under the manager folder.

Thanks

new/components/vcloud/update-vcloud-vm.sh
```diff
@@ -0,0 +1,22 @@
...
+
+uname -a
+
+# update system
+# yum update --skip-broken -y
+
+# install requirements for python-lxml
+yum install python-devel gcc-c++ gcc libxslt-devel libxml2-devel python-argparse zlib-devel -y
```
nir0s: @ran-z, @0lvin  I assume that we know this will not be production ready. If we want to be able to provide this in an offline method, we should compile whatever we need and create an rpm.

new/simple-manager-blueprint.yaml
```diff
@@ -328,6 +331,7 @@ node_templates:
     properties:
       ssh_user: { get_input: ssh_user }
       ssh_key_filename: { get_input: ssh_key_filename }
+      agent_local_key_path: { get_input: agent_private_key_path }
```
nir0s: why do we need this for simple?

new/vcloud-manager-blueprint.yaml
```diff
@@ -0,0 +1,867 @@
...
+
+
+######################################################################
+# This is a simple blueprint specific node_type to allow us to define
+# The public_ip of the machine is an accessible property of the host.
+# By default, the `ip` property is the private ip.
+######################################################################
+node_types:
+
+  vcloud_configuration:
```
nir0s: this is not consistent with the aws/openstack blueprint. In openstack, we use dsl_definitions and in openstack we use a node_template for the config. Can we choose either? I guess dsl_definitions are cleaner and should apply this to aws as well. (@ChenRoth)

new/vcloud-manager-blueprint.yaml
```diff
@@ -0,0 +1,867 @@
...
+
+  manager_host_server:
+    type: cloudify.vcloud.nodes.Server
+    properties:
+      server:
+        name: { get_input: server_name }
+        catalog: { get_input: catalog }
+        template: { get_input: template }
+        hardware:
+          cpu: 3
```
nir0s: should this (and memory) not be an input?

new/vcloud-manager-blueprint.yaml
```diff
@@ -0,0 +1,867 @@
...
+          memory: 4096
+        guest_customization:
+          computer_name: { get_input: server_name }
+      install_agent: false
+      management_network: { get_input: management_network_name }
+      vcloud_config: { get_property: [vcloud_configuration, vcloud_config] }
+    relationships:
+      #- target: management_network
```
nir0s: can we remove these comments?

new/vcloud-manager-blueprint.yaml
```diff
@@ -0,0 +1,867 @@
...
+      - target: manager_public_ip
+        type: cloudify.vcloud.server_connected_to_public_nat
+      - target: ssh_keypair
+        type: cloudify.vcloud.server_connected_to_keypair
+      - target: management_port
+        type: cloudify.vcloud.server_connected_to_port
+      - target: manager_security_group
+        type: cloudify.vcloud.server_connected_to_security_group
+
+  manager_host:
```
nir0s: just wondering.. why do we need both a manager_host_server node and a manager_host one?

new/vcloud-manager-blueprint.yaml
```diff
@@ -0,0 +1,867 @@
...
+    interfaces:
+      cloudify.interfaces.lifecycle:
+        create:
+          implementation: fabric.fabric_plugin.tasks.run_script
+          inputs:
+            script_path: components/manager/scripts/configure_manager.sh
+            fabric_env:
+              user: { get_input: ssh_user }
+              key_filename: { get_input: ssh_key_filename }
+              host_string: { get_attribute: [manager_public_ip, public_ip] }
```
nir0s: I see that in the manager_host node you also provide an ssh port to the fabric_env. As this node relates directly to the manager_host, shouldn't we apply the same port here?
 
Moreover, you can use yaml aliases instead of redefining the fabric env dict.

```yaml
fabric_env: &my_fabric_env
   host_string: ...
   ... 

...
fabric_env: *my_fabric_env
```

new/vcloud-manager-blueprint.yaml.template
```diff
@@ -0,0 +1,19 @@
+management_network_name: some_managements_network
```
nir0s: @0lvin @ran-z don't we want to provide the same inputs file format with documentation and all?
Also, can we please rename this file to correspond with the current convention? (i.e. vcloud-manager-blueprint-inputs.yaml)

new/simple-manager-blueprint.yaml
0lvin: Without this property after install manager we don't have private part of key for agent. I delete it  if it does not needed for simple install.

nir0s: Well, while we might decide we need it, please remove it for now and we'll add it if it becomes relevant.

new/vcloud-manager-blueprint.yaml
0lvin: cool, i will use it

new/components/vcloud/update-vcloud-vm.sh
0lvin: Do we can add python-lxml to default cloudify virtual environment that used for create deployments? If it can be true we does not need install compilers at all.

nir0s: Well, I guess this is needed for the vCloud api python package itself
right? If so, once we're done with the offline plugins feature, this won't
be needed.

new/vcloud-manager-blueprint.yaml
0lvin: manager_host_server - respond for creating vapp in cloud.
manager_host - respond for update and configure this server.

nir0s: Well, manager_configuration is meant to be a node in which we dump configuration and runtime properties for others to read and use, so we probably don't want to merge them.

What I meant is, that we can merge manager_host_server and manager_host (like in the openstack manager blueprint). I fail to understand why we need both.

new/vcloud-manager-blueprint.yaml
```diff
@@ -0,0 +1,876 @@
...
+            fabric_env: &simple_fabric_env
+              user: { get_input: ssh_user }
+              key_filename: { get_input: ssh_key_filename }
+              host_string: { get_attribute: [manager_public_ip, public_ip] }
+              port: { get_input: ssh_port }
+              disable_known_hosts: True
+        configure:
+          implementation: fabric.fabric_plugin.tasks.run_task
+          inputs:
+            tasks_file: components/manager/scripts/vcloud/configure.py
```
nir0s: um, why is configure being run here instead of as part of the manager_configuration node? Again, you can take an example from the openstack manager blueprint. The configure script sets the provider context for the manager which is unrelated to its physical host but rather to its config.

new/vcloud-manager-blueprint-inputs.yaml
```diff
@@ -0,0 +1,19 @@
+management_network_name: some_managements_network
```
nir0s: Can you please format this file as it is formatted in the openstack/simple manager blueprints?

new/vcloud-manager-blueprint.yaml
```diff
@@ -0,0 +1,876 @@
...
+  user_public_key:
+    type: string
+    description: >
+      Public part of your key
+
+  management_port_ip_allocation_mode:
+    type: string
+    default: dhcp
+    description: >
+      Ip allocation type for case when you doesn't set public ip
```
nir0s: 'IP allocation type for cases when you don't set a public ip for a node and instead want to auto-allocate it.'

new/vcloud-manager-blueprint.yaml
```diff
@@ -0,0 +1,876 @@
...
+    type: string
+    default: 2
+    description: >
+      Count cpu on node
+
+  manager_server_memory:
+    type: string
+    default: 4096
+    description: >
+      Amount memmory on node
```
nir0s: 'Manager host memory allocation'

new/vcloud-manager-blueprint.yaml
```diff
@@ -0,0 +1,876 @@
...
+
+  #############################
+  # Provider specific Inputs
+  #############################
+  public_ip:
+    type: string
+    default:
+    description: >
+      Public ip, can be empty for auto allocate
```
nir0s: allocate -> allocation.

new/vcloud-manager-blueprint.yaml
nir0s: ?

0lvin: i see, have fixed

PR #142, opened 2015-10-11
CFY-3459 add an import resolver section to the new manager blueprints

yaelpeled:


new/openstack-manager-blueprint.yaml
```diff
@@ -426,6 +426,14 @@ node_templates:
         policy_engine:
           start_timeout: 30
 
+        import_resolver:
```
nir0s: please add documentation explaining what this means... including the fact that it fails over to the online files, etc..

PR #157, opened 2015-10-24
CFY-3224-Make-rabbit-security-work-with-ctx-j

geokala:


new/components/amqpinflux/scripts/create.sh
```diff
@@ -29,7 +29,7 @@ yum_install ${AMQPINFLUX_RPM_SOURCE_URL}
 
 create_service_user ${AMQPINFLUX_USER} ${AMQPINFLUX_HOME}
 
-if [[ "${RABBITMQ_CERT_ENABLED}" == 'True' ]]; then
+if [[ "${RABBITMQ_CERT_ENABLED}" == 'true' ]]; then
```
nir0s: Are you sure this would work? This isn't a boolean...

geokala: It does work- tested.

The 'ctx -j node properties xyz' retrieves the named property as the json value of true (lower case). If not using '-j' then we get the python version but nothing when it's false.

PR #158, opened 2015-10-25
CFY-3901 bootstrap sanity test impl

idanmo:


tests/requirements.txt
```diff
@@ -0,0 +1,3 @@
+retrying
```
dankilman: maybe we should call this file - bootstrap-sanity-requirements.txt to make it clearer what it relates to

tests/sanity.py
```diff
@@ -0,0 +1,194 @@
...
+import os
+
+import boto.ec2
+from retrying import retry
+from fabric.api import settings as fabric_settings
+from fabric.api import run as fabric_run
+
+
+REGION = 'us-east-1'
+CENTOS_7_AMI_ID = 'ami-96a818fe'
```
dankilman: we are logging this ami public ip, are we certain it does not have password ssh enabled?

tests/sanity.py
```diff
@@ -0,0 +1,194 @@
...
+lgr.setLevel(logging.INFO)
+
+aws_access_key = os.environ['AWS_ACCESS_KEY']
+aws_secret_access_key = os.environ['AWS_SECRET_ACCESS_KEY']
+ssh_public_key = os.environ['SSH_PUBLIC_KEY']
+ssh_private_key = os.environ['SSH_PRIVATE_KEY']
+
+
+def delete_security_group(conn):
+    sgs = [
```
dankilman: regarding all the `get_all_XXXX`, is there any server side filter to avoid filtering on the client?
there could be many keypairs/security_groups/instances.

PR #168, opened 2015-11-06
CFY-3223-allow-external-rabbit

geokala:


new/components/rabbitmq/scripts/create.sh
```diff
@@ -4,92 +4,108 @@
 
 CONFIG_REL_PATH="components/rabbitmq/config"
 
-export ERLANG_SOURCE_URL=$(ctx node properties erlang_rpm_source_url)
-export RABBITMQ_SOURCE_URL=$(ctx node properties rabbitmq_rpm_source_url)
-export RABBITMQ_FD_LIMIT=$(ctx node properties rabbitmq_fd_limit)
+export RABBITMQ_USE_EXTERNAL=$(ctx -j node properties rabbitmq_use_external)
```
nir0s: Well, since external influx and external elasticsearch are already implemented in such a way that all you have to do is provide an endpoint that isn't an empty string as an input to the blueprint, and you don't have to explicitly use another "use external" input, can we do the same here? If we want to be more explicit, we can do that in 3.4.

sounds ok?

geokala: That's fair enough- I was following the rabbit SSL paradigm but that's simple enough as there's only two places it's actually used (the rabbit scripts).

PR #198, opened 2015-11-24
cleanup inputs for vsphere and add private vcloud input

0lvin:
Add additional inputs descriptions for vsphere
Add input for support token based login for private vcloud (vcloud_org_url)

### DIFF
--- a/vcloud-manager-blueprint-inputs.yaml
+++ b/vcloud-manager-blueprint-inputs.yaml
@@ -22,6 +22,10 @@ vcloud_password: ''
 # Organization name
 vcloud_org: ''
 
+# Vcloud org url, only required if using token based login on a private
+# vcloud director.
+#vcloud_org_url: ''
+
 # Virtual data center name
 vcloud_vdc: ''
 
--- a/vcloud-manager-blueprint.yaml
+++ b/vcloud-manager-blueprint.yaml
@@ -128,6 +128,15 @@ inputs:
     description: >
       Organization name
 
+  vcloud_org_url:
+    type: string
+    default: ''
+    description: >
+      Only required if using token based login on a private vcloud
+      director. This can be obtained by following the vcloud API
+      example docs. If you are unsure of how to obtain this, you will
+      need to use password based login.
+
   vcloud_instance:
     type: string
     default: ''
@@ -585,6 +594,7 @@ node_templates:
         service_type: { get_input: vcloud_service_type }
         service: { get_input: vcloud_service }
         api_version: { get_input: vcloud_api_version }
+        org_url: { get_input: vcloud_org_url }
         edge_gateway: { get_input: edge_gateway }
         ssl_verify: { get_input: ssl_verify }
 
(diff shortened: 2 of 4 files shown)


vsphere-RHEL-manager-blueprint-inputs.yaml
```diff
@@ -9,15 +9,15 @@
 #vsphere_port: 443
 #vsphere_datacenter_name: Datacenter
 #vsphere_resource_pool_name: Resources
-#vsphere_auto_placement: true
+#vsphere_auto_placement: false
```
nir0s: just wondering.. are these changes from "true" to "false" and the likes related to the fact that they were wrong before?

vsphere-RHEL-manager-blueprint-inputs.yaml
```diff
@@ -9,15 +9,15 @@
...
 # Template
-#template: "CentOS7_1406_tmpl"
+#template: "CentOS-7-x86_64-1406"
 
 # Names of components
 #management_network_name: "Management"
-#external_network_name: "Application"
-#server_name: "cloudify-manager-server"
+#external_network_name: "DMZ"
+#server_name: "centosmanager"
```
nir0s: why change the name of the server? The previous one actually makes more sense to me.

vsphere-RHEL-manager-blueprint-inputs.yaml
```diff
@@ -9,15 +9,15 @@
 #vsphere_port: 443
 #vsphere_datacenter_name: Datacenter
 #vsphere_resource_pool_name: Resources
-#vsphere_auto_placement: true
+#vsphere_auto_placement: false
 
 # Template
-#template: "CentOS7_1406_tmpl"
+#template: "CentOS-7-x86_64-1406"
```
nir0s: What does template mean? In general, what we want to achieve here is that each and every configuration property is documented enough so that a user will need to do as little searching as possible to answer what each property means.

vsphere-RHEL-manager-blueprint-inputs.yaml
```diff
@@ -26,31 +26,25 @@
...
 #ssh_key_filename: ~/.ssh/centos_key.pem
+# SSH key path used to connect from the manager to nodes
 #agent_private_key_path: ~/.ssh/cloudify-agent-kp.pem
 
-# SSH port used to connect to the manager
-#ssh_port: 22
-
 # This is the user with which the Manager will try to connect to the application hosts.
-#agents_user: ''
+#agents_user: 'ubuntu'
```
nir0s: I think this should actually not have a default as it very much depends on the user's stack.

vsphere-RHEL-manager-blueprint-inputs.yaml
0lvin: not wrong, in blueprint we have different value

0lvin: i have added little description for this field

nir0s:
please resolve conflicts. Also, please add a description for every input. I see that some inputs simply don't have a description in the inputs file which means that a user will have to guess what they mean.

Thanks.

0lvin:
Updated, please review

PR #199, opened 2015-11-25
cleanup inputs for vcloud

kostya13:


nir0s:
um.. tests aren't passing

vcloud-RHEL-manager-blueprint-inputs.yaml
```diff
@@ -1,5 +1,5 @@
 #############################
-# vCloud Air specific Inputs
+# vCloud specific Inputs
```
nir0s: the idea was that each input in the input files will have an explanation and that WHATEVER can have a default, will have a default so that users can configure as less as possible before bootstrapping.
Please add the descriptions and defaults where possible.
Thanks

kostya13:
Travis CI - passed.

vcloud-RHEL-manager-blueprint-inputs.yaml
kostya13: I've added comments for all parameters.

PR #215, opened 2015-12-09
CFY-4382 agents sanity checks passed through nginx

konradxyz:


dankilman:
not relevant, endpoint is removed from other PRs

PR #221, opened 2015-12-28
CFY-3717 copy roles mapping to manager on bootstrap

adamlavie:


components/restservice/scripts/install_plugins.sh
```diff
@@ -7,20 +7,24 @@ echo "Installing rest service plugins"
...
-plugin_dir=$(python -c "${extract_plugin_script}")
+    plugin_dir=$(python -c "${extract_plugin_script}")
 
-pushd ${plugin_dir}
-sudo $VENV/bin/pip install . {{ plugin.install_args }}
-popd
+    pushd ${plugin_dir}
+        if [[ {{ plugin.source }} == *.wgn ]]; then
+            # install wagon plugin
+            sudo $VENV/bin/pip install {{ plugin.install_args }}
```
noakup: do we really need sudo to run pip install inside a virtualenv?

components/restservice/scripts/install_plugins.sh
```diff
@@ -7,20 +7,24 @@ echo "Installing rest service plugins"
...
+        else
+            sudo $VENV/bin/pip install . {{ plugin.install_args }}
```
noakup: same

components/restservice/scripts/install_plugins.sh
adamlavie: We do, actually. It's required to use sudo to install dependencies in the manager env

components/restservice/scripts/install_plugins.sh
adamlavie: same..

PR #228, opened 2016-01-06
Convert rabbitmq

geokala:
Also reactivated all of the components so that I can test deployments locally.


components/utils.py
```diff
@@ -118,16 +128,17 @@ def sudo(command):
...
 
 
-def create_dir(dir):
-    if os.path.isdir(dir):
+def create_dir(directory):
+    if os.path.isdir(directory):
         return
-    logger.info('Creating Directory: {0}'.format(dir))
-    sudo(['mkdir', '-p', dir])
+    else:
```
nir0s: don't quite understand the "else" here...

components/utils.py
```diff
@@ -334,6 +345,10 @@ def start(service_name):
     def stop(service_name):
         sudo(['systemctl', 'stop', service_name])
 
+    @staticmethod
+    def reload_configuration():
```
nir0s: can we call this "daemon_reload" to be consistent with systemd's naming?

components/utils.py
```diff
@@ -342,6 +357,27 @@ def move(source, destination):
...
+    # The python equivalent to this is an ugly combination of os.walk and
+    # os.chown.
+    # Leaving group blank will use the default chown behaviour of setting the
+    # group that is the user's default
+    owner_string = '{owner}.{group}'.format(
+        owner=user,
+        group=group,
+    )
+
+    run(['sudo', 'chown', '-R', owner_string, path])
```
nir0s: you can use `sudo` isntead of `run` which appends `sudo` instead..

components/utils.py
```diff
@@ -407,73 +443,79 @@ def chown(user, group, path):
...
+#     # Remove the directory if it's empty, ignoring failures due to lack of directory  # noqa
+#     # This won't remove /var/log if ${service} is empty, unless /var/log is empty.  # noqa
+#     # It will, however, error if its target dir is non-empty
+
+#     ctx logger info "Removing unnecessary logs directory: /var/log/${service}"  # noqa
+#     sudo rm -df /var/log/${service}
 # }
 
+
+def run_with_retries(command, retries=5, delay=3):
```
nir0s: can you please incorporate this within `run` instead?

components/rabbitmq/scripts/create.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+config_path = "components/rabbitmq/config"
+
+rabbitmq_endpoint_ip = ctx.node.properties('rabbitmq_endpoint_ip')
+
+
+def install_rabbitmq():
+    # Sources
+    erlang_source_url = ctx.node.properties('erlang_rpm_source_url')
```
nir0s: while I agree that we should definitely come up with something cleverer than having all vars as globals, I think we should at least be consistent in the beginning.

please move all service definitions (paths, ports, ctx properties, etc..) to be global at the top and we'll come up with a good solution later on.

See influx as reference. thanks

components/utils.py
```diff
@@ -342,6 +357,27 @@ def move(source, destination):
...
 
+def set_permissions_owner_execute_only(path):
+    os.chmod(
+        path,
+        # Set read and execute for owner
+        stat.S_IRUSR | stat.S_IXUSR,
+    )
+
+
+def set_ownership_recursive(user, path, group=''):
```
nir0s: please use this: https://github.com/cloudify-cosmo/cloudify-manager-blueprints/pull/228/files#diff-23cb882db3b8314e49ad99212e1c4118R441

components/utils.py
geokala: I added it when I was tweaking the name to make the logic flow more explicit- either the directory exists OR we create it.

Also means that we don't accidentally try to create it if we decide that there needs to be an extra line of logging, it won't cause problems if whoever implements it brainos and deletes the return.

components/utils.py
geokala: Will do- I named it like this because it actually makes more sense to anyone who doesn't work with systemd much. I'll add a comment to make it clear for anyone in that camp.

components/utils.py
geokala: That would've been smart- sorting that now.

PR #230, opened 2016-01-20
CFY-4762 - customizable management workers logging level

isaac-s:


dankilman:
PR looks good.
Please restore the default back to `debug` and open a JIRA to consider changing it to `info`.
It should be noted that in 3.4, only celery specific logs will go to the celery.logs. All application/cloudify/etc... logs will go to to a different place.

isaac-s:
Default changed back to `debug` in manager blueprints.

The (commented-out) setting in the input templates remained as `info`. I'd like to keep it like that as some sort of a "hint" (unless we have a convention whereby the inputs template must contain the defaults commented out).

Separate JIRA for considering `info` as the default: https://cloudifysource.atlassian.net/browse/CFY-4781

dankilman:
I would prefer it if we restore the commented out part to "debug" as well

dankilman:
also, please squash your commits after you are done

nir0s:
Isaac, please provide a very clear description of what this does. While it is obvious to us, users who will be looking at the inputs file have no idea what the management worker is so they will not know what the log level will be used for.

Thanks.

isaac-s:
@nir0s done.

nir0s:
Sorry. Again? :)

PR #234, opened 2016-01-27
CFY-4814: additional use_external_resource

isaac-s:
Allow user to use external resources for additional resources.

Changes added only for OpenStack at the moment. If the approach is acceptable, let me know and I will replicate for other clouds.


nir0s:
I think this needs a wider audience and should be brought up for discussion.

PR #237, opened 2016-02-03
CFY-4854 - removed duplicate inputs

isaac-s:


nir0s:
only relevnt in simple?

isaac-s:
@nir0s yes.

PR #251, opened 2016-02-29
CFY-4901: Ensure envvars are exported if security is enabled

isaac-s:


components/manager/scripts/validate.py
```diff
@@ -0,0 +1,20 @@
+from cloudify import ctx
+from cloudify_cli import constants
+
+from cloudify.exceptions import NonRecoverableError
+import os
+
+try:
+    security_enabled = ctx.node.properties['security']['enabled']
+except KeyError:
+    security_enabled = False
```
max-orlov: maybe security_enabled = ctx.node.properties.get('security', {}).get('enabled') ?

components/manager/scripts/validate.py
```diff
@@ -0,0 +1,20 @@
+from cloudify import ctx
+from cloudify_cli import constants
+
+from cloudify.exceptions import NonRecoverableError
+import os
```
max-orlov: on top of the import stack

components/manager/scripts/validate.py
ran-z: For achieving the same intended result,
security_enabled = ctx.node.properties.get('security', {}).get('enabled', False)

components/manager/scripts/validate.py
```diff
@@ -0,0 +1,20 @@
...
+
+if security_enabled:
+    missing_keys = {constants.CLOUDIFY_USERNAME_ENV,
+                    constants.CLOUDIFY_PASSWORD_ENV} -\
+                   set(os.environ.keys())
+
+    if missing_keys:
+        raise NonRecoverableError(
+            'Security is enabled, but the following required environment '
+            'variables have not been set: {}'.format(list(missing_keys)))
```
ran-z: 1) use either "{0}" or "{<name>}", not "{}" (non standard)
2) printing as a list would have "[" and "]" in the output, im not sure if that was your intended result there, i'd probably have it without them brackets :)

isaac-s:
@ran-z @max-orlov 
Fixed, please re-check

components/manager/scripts/validate.py
```diff
@@ -0,0 +1,17 @@
+import os
```
max-orlov: Though the functionality is rather the same, perhaps is should be implemented in the cli code rather than in a blueprint. Since the bootstrap process itself uses the rest client regardless of the client usage.

isaac-s: We have tons of other operations done in the blueprint rather than in the CLI. We should have some sort of a criteria what type of functionality exists in the UI and what's in the blueprint. For validation, I'd prefer the blueprint because I don't want to depend on the CLI to perform validation — maybe in the future we'd like to allow bootstrapping though means other than the CLI.

max-orlov: Yeah, that could very well be the case in the future, but while it is a good idea to validate in the blueprint, the CLI cannot rely that such validation indeed occur.

ran-z: If we'll want to allow bootstrapping outside of the CLI, we'll need to make several adjustments to the code as it is.

the relevant question is whether the user/pass are needed as part of the workflow code or as part of the CLI code that runs during bootstrap. at the moment, it's the latter (or possibly, both). therefore, the validation should be as part of the CLI, so in case a user uses a different workflow, the validation will still happen.
in theory, if we allow bootstrapping via other means than the CLI, then the validation will *also* need to happen as part of the workflow - but not instead of the one in the CLI.
at the moment, i think the best is to go for a single validation in the CLI, and not do a double validation for something that might or might not happen in the future.

isaac-s:
I'm convinced. I'll close this PR.

PR #259, opened 2016-03-24
Save logs and events to file.

kostya13:
Save cloudify events and log to files.
This feature uses together with Log Insight agent.


nir0s:
As we discussed. This is environment specific. We require a generic solution here.

PR #260, opened 2016-03-25
CFY-5108: added inputs for manager configuration

isaac-s:


idanmo:
Thanks, we'll address this in a wider manner for all inputs we would like to expose.

PR #267, opened 2016-03-29
CFY-4936 added maintenance endpoint and conf

nirbir:


components/restservice/config/cloudify-rest.conf
```diff
@@ -17,5 +17,6 @@
     file_server_uploaded_blueprints_folder: 'uploaded-blueprints',
     file_server_snapshots_folder: 'snapshots',
     file_server_resources_uri: '/resources',
-    insecure_endpoints_disabled: {{ ctx.node.properties.insecure_endpoints_disabled }}
+    insecure_endpoints_disabled: {{ ctx.node.properties.insecure_endpoints_disabled }},
+    maintenance_folder: '/opt/manager/maintenance'
```
nir0s: um... don't we need to add this to the rest service's create.py script so that it creates the dir?

nirbir: I create this folder in the rest service when activating maintenance mode

nir0s: alright.

PR #275, opened 2016-04-10
CFY-4952 upgrade support ctx properties

adamlavie:


components/amqpinflux/scripts/create.py
```diff
@@ -67,7 +69,7 @@ def install_amqpinflux():
     ctx.logger.info('Fixing permissions...')
     utils.chown(amqpinflux_user, amqpinflux_group, AMQPINFLUX_HOME)
 
-    utils.systemd.configure('amqpinflux')
```
adamlavie: pass only the service name and normalize to be consistent with rest of scripts

components/elasticsearch/config/elasticsearch.yml
```diff
@@ -1,8 +1,8 @@
 ############################ Cloudify Configuration ############################
 
-http.port: {{ ctx.node.properties.es_endpoint_port }}
-discovery.zen.ping.multicast.enabled: {{ ctx.node.properties.es_clustering_enabled }}
-discovery.zen.ping.multicast.port: {{ ctx.node.properties.es_clustering_discovery_port }}
```
adamlavie: make nested inside properties/node-properties

components/utils.py
```diff
@@ -217,16 +224,35 @@ def download_cloudify_resource(url):
     return destf
 
 
-def deploy_blueprint_resource(source, destination):
+# User resources are resources non related to the cloudify core configuration
+# such as security userstore and security config files and ssl certificates.
+# These files will be handled differently then config files as they will remain
+def deploy_user_resource(source, destination, params):
+    ctx.logger.info('Deploying user resource {0} to {1}'.format(
+            source, destination))
```
adamlavie: Create the factory once

components/utils.py
```diff
@@ -217,16 +224,35 @@ def download_cloudify_resource(url):
...
     ctx.logger.info('Deploying blueprint resource {0} to {1}'.format(
         source, destination))
-    tmp_file = ctx.download_resource_and_render(source)
-    move(tmp_file, destination)
+    resource_file, dest = BlueprintResourceFactory().create(source,
+                                                            destination,
+                                                            params)
+    copy(resource_file, dest)
+
+
```
adamlavie: no one is using

components/utils.py
```diff
@@ -310,26 +336,72 @@ def yum_install(source):
         source_name = subprocess.check_output(
             ['rpm', '-qp', source_path]).strip()
 
-    ctx.logger.info('Checking whether {0} is already installed...'.format(
-        source_path.rstrip('\n\r')))
-    installed = run(['rpm', '-q', source_name], ignore_failures=True)
-    if installed.returncode == 0:
-        ctx.logger.info('Package {0} is already installed.'.format(source))
-        return
```
adamlavie: unindent

components/utils.py
```diff
@@ -421,12 +497,20 @@ def set_selinux_permissive():
         ctx.logger.info('SELinux is not enforced.')
 
 
-def get_rabbitmq_endpoint_ip():
+def set_rabbitmq_policy(name, q_regex, p_type, value):
```
nir0s: this has been changed and is not rebased on top of master.

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
     ctx.logger.debug('Extracting {0} to {1}...'.format(source, destination))
     sudo(['tar', '-xzvf', source, '-C', destination,
           '--strip={0}'.format(strip)])
+
+
+def write_file(content, file_path):
+    tmp_file = tempfile.NamedTemporaryFile(delete=False)
+    mkdir(os.path.dirname(os.path.abspath(file_path)))
```
nir0s: I wouldn't add that here. this function explicitly writes a file.. it's called write_file. I would create a directory before running the write_file outside of this function.

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+    with open(tmp_file.name, 'w') as f:
+        f.write(json.dumps(content))
```
nir0s: what if it's not json?

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+    move(tmp_file.name, file_path)
+
+
+class CtxPropertyFactory(object):
+    PROPERTIES_FILE_NAME = 'properties.json'
+    PROPERTIES_PATH = '/opt/cloudify/node_properties/'
```
nir0s: I don't quite understand why we're separating these according to `node_properties`, etc.. and not according to a service.
I think it would make much more sense to create something like `/opt/cloudify/amqpinflux/node_properties`, `/opt/cloudify/amqpinflux/resources`, etc.. for each service which would make handling each service easier. When we move to a distributed manager, the handling of each upgrade mechanism would have to be on the node level.

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+    # A list of property suffixes to be included in the upgrade process,
+    # despite having 'use_existing_on_upgrade' set to ture
+    UPGRADE_PROPS_SUFFIX = ['source_url', 'cloudify_resources_url',
+                            'use_existing_on_upgrade']
+
+    def __init__(self):
+        self.is_upgrade = is_upgrade()
+
+    # Create node properties according to the workflow context install/upgrade
+    def create(self, service_name, write_to_file=True):
```
nir0s: I don't quite understand why we need "Factories" here rather than just regular "Classes". Factories are used to generate objects dynamically based on different parameters whereas here our factories do something completely different than one another.

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+
+def write_file(content, file_path):
+    tmp_file = tempfile.NamedTemporaryFile(delete=False)
+    mkdir(os.path.dirname(os.path.abspath(file_path)))
+    with open(tmp_file.name, 'w') as f:
+        f.write(json.dumps(content))
+    move(tmp_file.name, file_path)
+
+
+class CtxPropertyFactory(object):
```
nir0s: I don't quite understand why we need "Factories" here rather than just regular "Classes". Factories are used to generate objects dynamically based on different parameters whereas here our factories do something completely different than one another.

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+            dest_file_path = os.path.join(
+                    base_service_dir,
+                    'upgrade-{0}'.format(self.PROPERTIES_FILE_NAME))
+        else:
+            dest_file_path = os.path.join(base_service_dir,
+                                          self.PROPERTIES_FILE_NAME)
+        return dest_file_path
+
+    def _load_ctx_properties(self, service_name):
+        node_props = dict(ctx.node.properties.get_all())
```
nir0s: is there a situation in which this isn't a dict?

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+        if self.is_upgrade:
+            # Use existing property configuration during upgrade
+            use_existing = node_props.get('use_existing_on_upgrade')
+            if use_existing:
+                install_props_path = os.path.join(self.PROPERTIES_PATH,
+                                                  service_name,
+                                                  self.PROPERTIES_FILE_NAME)
+                ctx.logger.info('Loading existing {service} input config from '
+                                '{path}'.format(service=service_name,
```
nir0s: while I agree that this is more readable (i.e. it looks more like english than {0}), the convention we're using across all scripts is not to explicitly provide the name of the variables. Is  there a reason to break consistency here?

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+                         rollback_dir)
+
+            for filename in resource_files:
+                if filename.startswith(self.UPGRADE_RESOURCE_PREFIX):
+                    move(os.path.join(service_archive_path, filename),
+                         os.path.join(service_archive_path, filename[8:]))
+
+
+def is_upgrade():
+    '''
```
nir0s: for all docstrings, again, the convention we use here is:

```python
"""first sentence

bla bla bla
"""

can we not break it?

File: components/utils.py [hunk 15]
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+    Returns true if manager is in upgrade mode. This function will assume
+    manager is in upgrade state according to the maintenance mode file.
+    :return: True if in upgrade state, false otherwise.
+    '''
+    status_file_path = '/opt/manager/status.txt'
+    if os.path.isfile(status_file_path):
+        ctx.logger.info('loading upgrade status file: {0}'
```
nir0s: capital L please

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+            status_dict = json.load(f)
+        return status_dict['status'] == 'activated'
+    else:
+        return False
+
+
+def is_rollback():
+    '''
+    Returns true if manager is in rollback state.
+    :return: True if in rollback state, false otherwise.
```
nir0s: same for docs stuff

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+    '''
+    return False
+
+
+def start_service_and_archive_properties(service_name,
```
nir0s: the name of this function is misleading. using it I would think that it does exactly what it says. thing is.. if I use it outside of an upgrade context, it would just start a service. If I use it inside an upgrade context, it doesn't just archive properties and restart.. it restarts a service and archives both properties and resources. I suggest we make the name more generic.

components/utils.py
```diff
@@ -310,26 +336,72 @@ def yum_install(source):
...
+            return
+
+        if is_upgrade() or is_rollback():
+            _remove_existing_rpm_package(source_name)
 
     ctx.logger.info('yum installing {0}...'.format(source_path))
     sudo(['yum', 'install', '-y', source_path])
 
 
+def _remove_existing_rpm_package(source_name):
```
nir0s: considering the set of functions we use to handle rpm files, i'd create a class called `RpmHandler` and instantiate it once from the yum_install function to keep the code clean.

components/utils.py
```diff
@@ -310,26 +336,72 @@ def yum_install(source):
...
+    package_name, package_version = _get_rpm_package_details(source_name)
+    existing_src_name = _get_installed_rpm_source(package_name)
+    if existing_src_name:
+        existing_package_name, existing_pkg_version = \
+            _get_rpm_package_details(existing_src_name)
+        ctx.logger.info('Version {0} is currently installed for package {1}. '
+                        'Uninstalling to allow upgrade to version {2}'
+                        .format(existing_pkg_version, existing_src_name,
+                                package_version))
+        sudo(['rpm', '--noscripts', '-e', existing_src_name])
```
nir0s: `yum update package.rpm` implements updating a package. Don't think we need to uninstall it first. you can also use `rpm -Uvh package.rpm` if needed.

components/utils.py
```diff
@@ -310,26 +336,72 @@ def yum_install(source):
...
-        return
+        ctx.logger.info('Checking whether {0} is already installed...'.format(
+            source_path))
+        installed_rpm = _get_installed_rpm_source(source_name)
+        if installed_rpm == source_name:
+            ctx.logger.info('Package {0} is already installed.'.format(source))
+            return
+
+        if is_upgrade() or is_rollback():
+            _remove_existing_rpm_package(source_name)
```
nir0s: `yum update package.rpm` implements updating a package. Don't think we need to uninstall it first. you can also use `rpm -Uvh package.rpm` if needed.

instead if we call the relevant yum subcommand according to the context (bootstrap vs. upgrade)

components/utils.py
```diff
@@ -310,26 +336,72 @@ def yum_install(source):
...
+    installed = run(['rpm', '-q', name], ignore_failures=True)
+    if installed.returncode == 0:
+        existing_rpm_pkg_name = installed.aggr_stdout.rstrip('\n\r')
+        ctx.logger.debug('Found installed package named: {0}'
+                         .format(existing_rpm_pkg_name))
+        return existing_rpm_pkg_name
+    return None
+
+
+def _get_rpm_package_details(source_name):
```
nir0s: if a user puts the rpm within their yum repo instead of providing a url to a file, this will not work. a lot of customers want to put the rpms we provide within their yum repos and then just provide a name of the package in the blueprint.

Additionally, we can't assume that the name of the file is indicative of its version and real package name even if the rpm conventions declare so.

rpm/yum provide an interface (horrible one though robest) for getting the real package name and version (https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Deployment_Guide/sec-Displaying_Package_Information.html). This prevents ambiguities and future problems with identification.

components/utils.py
```diff
@@ -310,26 +336,72 @@ def yum_install(source):
...
+            _get_rpm_package_details(existing_src_name)
+        ctx.logger.info('Version {0} is currently installed for package {1}. '
+                        'Uninstalling to allow upgrade to version {2}'
+                        .format(existing_pkg_version, existing_src_name,
+                                package_version))
+        sudo(['rpm', '--noscripts', '-e', existing_src_name])
+
+
+def _get_installed_rpm_source(name):
+    '''
```
nir0s: why do we need to return the `existing_rpm_pkg_name` here? We must know the formal name beforehand to even check if it's installed in the first place? `yum_install` implemented that as well in L336

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+                            del existing_props[key]
+
+                # Update node properties with existing configuration inputs
+                node_props.update(existing_props)
+
+        node_props['service_name'] = service_name
+
+        return node_props
+
+
```
nir0s: between the two blueprint related classes, I don't see a place in which we backup what's in `/opt/cloudify/sources` which contains the relevant packages for the rollback. Am I missing something?

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+                else:
+                    ctx.logger.info('User resource {resource} not found on'
+                                    ' {path}'.format(resource=resource_name,
+                                                     path=dest))
+
+        if not os.path.isfile(dest):
+            self._download_resource(source, dest, params)
+
+    @staticmethod
+    def _download_resource(source, dest, params):
```
nir0s: how is this different than `deploy_blueprint_resource`

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+        resource_name = os.path.basename(dest)
+        ctx.logger.info('Downloading resource {resource} to '
+                        '{dest}'.format(resource=resource_name, dest=dest))
+        ctx.download_resource_and_render(source, dest, params)
+
+    def _get_resource_file_path(self, service_name, resource_name, upgrade):
+        base_service_dir = os.path.join(self.RESOURCES_PATH, service_name)
+        if upgrade:
+            upgrade_resource_name = '{0}{1}'\
```
nir0s: nothing big.. just.. I think it would be nicer to write `self.UPGRADE_RESOURCE_PREFIX + resource_name` here instead..

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+    @staticmethod
+    def _download_resource(source, dest, params):
+        resource_name = os.path.basename(dest)
+        ctx.logger.info('Downloading resource {resource} to '
+                        '{dest}'.format(resource=resource_name, dest=dest))
+        ctx.download_resource_and_render(source, dest, params)
+
+    def _get_resource_file_path(self, service_name, resource_name, upgrade):
+        base_service_dir = os.path.join(self.RESOURCES_PATH, service_name)
+        if upgrade:
```
nir0s: In general, I think we shouldn't propagate `update` anywhere and use it as a globally accessible context.
That means that when a script runs, it will know whether it's in the context of an upgrade or not and while importing utils there will already be a context for all of the functions. This will make the code less cumbersome and allow future features in the scripts to just use that context without having to explicitly pass `upgrade` everywhere.

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+                                                      upgrade)
+        write_file(resources_dict, resources_json)
+
+    def archive_resources(self, service_name):
+        rollback_dir = os.path.join(self.RESOURCES_ROLLBACK_PATH, service_name)
+        mkdir(rollback_dir)
+
+        service_archive_path = os.path.join(self.RESOURCES_PATH, service_name)
+        # nodes will not always use blueprint resources i.e. java/python
+        if os.path.isdir(service_archive_path) and self.is_upgrade:
```
nir0s: maybe do `if self.is_upgrade` first so that it doesn't check for the dir every time when it doesn't need to?

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+            ctx.logger.info('archiving node {node} properties'
```
nir0s: capital a please

components/amqpinflux/scripts/create.py
nir0s: Just to add on top of that for clarity's sake. What I meant is that the parameters can be generated dynamically according to which service you're currently on. This can happen in `download_resource_and_render` and anywhere else where the params are used.

components/webui/scripts/start.py
```diff
@@ -10,4 +10,5 @@
 import utils  # NOQA
 
 
-utils.systemd.start('cloudify-webui')
+ctx.logger.info('Starting WebUI Service...')
+utils.start_service_and_archive_properties('cloudify-webui')
```
nir0s: Just a reminder that once we normalize the service's name, we should also change all start/stop scripts to use those names

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+            resource_files = os.listdir(service_archive_path)
+            for filename in resource_files:
+                if not filename.startswith(self.UPGRADE_RESOURCE_PREFIX):
+                    move(os.path.join(service_archive_path, filename),
+                         rollback_dir)
+
+            for filename in resource_files:
+                if filename.startswith(self.UPGRADE_RESOURCE_PREFIX):
+                    move(os.path.join(service_archive_path, filename),
+                         os.path.join(service_archive_path, filename[8:]))
```
nir0s: I assume you want to remove the prefix.. in which case this implementation is bound to break. It should probably instead using the length of the prefix as a starting point.

components/utils.py
```diff
@@ -520,3 +606,282 @@ def untar(source, destination='/tmp', strip=1):
...
+
+
```
adamlavie: Call this function once and have the state loaded as part of the utils import

components/utils.py
```diff
@@ -263,7 +286,7 @@ def wait_for_port(port, host='localhost'):
     error_exit('Failed to connect to {0}:{1}...'.format(host, port))
 
 
-def yum_install(source):
```
adamlavie: change service name to named arg

components/utils.py
```diff
@@ -369,17 +445,39 @@ def get_service_file_path(service_name):
         sid = 'cloudify-{0}'.format(service_name)
         return "/usr/lib/systemd/system/{0}.service".format(sid)
 
-    def enable(self, service_name, retries=0):
-        ctx.logger.info('Enabling systemd service {0}...'.format(service_name))
```
adamlavie: append_prefix should be set to True by default

components/utils.py
```diff
@@ -520,3 +626,350 @@ def untar(source, destination='/tmp', strip=1):
     ctx.logger.debug('Extracting {0} to {1}...'.format(source, destination))
     sudo(['tar', '-xzvf', source, '-C', destination,
           '--strip={0}'.format(strip)])
+
+
+def write_to_json_file(content, file_path):
+    tmp_file = tempfile.NamedTemporaryFile(delete=False)
+    mkdir(os.path.dirname(os.path.abspath(file_path)))
+    with open(tmp_file.name, 'w') as f:
+        f.write(json.dumps(content))
```
nir0s: I think you're still getting this per invocation instead of once

components/utils.py
```diff
@@ -520,3 +626,350 @@ def untar(source, destination='/tmp', strip=1):
...
+        '''
+        service_archive_path = self._get_rollback_properties_dir(service_name)
+        mkdir(service_archive_path)
+        base_service_dir = self._get_properties_dir(service_name)
+        properties_file_path = os.path.join(base_service_dir,
+                                            self.PROPERTIES_FILE_NAME)
+        ctx.logger.info('Archiving previous inputs for service {service}'
+                        .format(service=service_name))
+        move(properties_file_path, service_archive_path)
+        ctx.logger.info('Setting new properties for service {service}'
```
nir0s: What about the {0} convention?

components/webui/scripts/stop.py
```diff
@@ -9,5 +9,8 @@
     join(dirname(__file__), 'utils.py'))
 import utils  # NOQA
 
+WEBUI_SERVICE_NAME = 'webui'
```
nir0s: why do we need this global variable in start/stop scripts?

components/utils.py
```diff
@@ -298,38 +328,89 @@ def yum_install(source):
...
+        rpm_handler.remove_existing_rpm_package(source_path)
+
     ctx.logger.info('yum installing {0}...'.format(source_path))
     sudo(['yum', 'install', '-y', source_path])
 
 
+class RpmPackageHandler(object):
+
+    def remove_existing_rpm_package(self, source_path):
+        '''
```
nir0s: The PEP8 convention for docstrings is:

```python
"""First, non-desciptive line.

Description

params
"""
```

Let's follow it please.

components/utils.py
```diff
@@ -141,6 +143,19 @@ def move(source, destination):
     sudo(['mv', source, destination])
 
 
+def copy(source, destination):
+    sudo(['cp', '-r', source, destination])
+
+
+def delete(path, ignore_failure=False):
+    if os.path.exists(path):
```
nir0s: can we call this "remove" please? to correspond with the actual command..

components/restservice/scripts/create.py
```diff
@@ -99,63 +104,72 @@ def _configure_dbus(rest_venv):
     dbus_glib_bindings = os.path.join('/usr', site_packages,
                                       '_dbus_glib_bindings.so')
     dbus_bindings = os.path.join('/usr', site_packages, '_dbus_bindings.so')
+    ctx.logger.info('if os.path.isdir(dbuslib) {0}'.format(
```
nir0s: ?

components/utils.py
```diff
@@ -520,3 +620,349 @@ def untar(source, destination='/tmp', strip=1):
...
+
+
+def write_to_json_file(content, file_path):
+    tmp_file = tempfile.NamedTemporaryFile(delete=False)
+    mkdir(os.path.dirname(os.path.abspath(file_path)))
+    with open(tmp_file.name, 'w') as f:
+        f.write(json.dumps(content))
+    move(tmp_file.name, file_path)
+
+
```
adamlavie: this should be called once!

components/utils.py
```diff
@@ -520,3 +620,361 @@ def untar(source, destination='/tmp', strip=1):
...
+    BASE_PROPERTIES_PATH = '/opt/cloudify'
+    NODE_PROPS_DIR_NAME = 'node_properties'
+    ROLLBACK_NODE_PROPS_DIR_NAME = 'node_properties_rollback'
+
+    # A list of property suffixes to be included in the upgrade process,
+    # despite having 'use_existing_on_upgrade' set to ture
+    UPGRADE_PROPS_SUFFIX = ['source_url', 'cloudify_resources_url',
+                            'use_existing_on_upgrade']
+
+    def __init__(self):
```
adamlavie: make global

components/utils.py
```diff
@@ -520,3 +620,361 @@ def untar(source, destination='/tmp', strip=1):
...
+class BlueprintResourceFactory(object):
+
+    BASE_RESOURCES_PATH = '/opt/cloudify'
+    RESOURCES_DIR_NAME = 'resources'
+    RESOURCES_ROLLBACK_DIR_NAME = 'resources_rollback'
+    RESOURCES_JSON_FILE = '__resources.json'
+
+    UPGRADE_RESOURCE_PREFIX = 'upgrade-'
+
+    def __init__(self):
```
adamlavie: This is global

components/utils.py
```diff
@@ -520,3 +620,361 @@ def untar(source, destination='/tmp', strip=1):
...
+        """
+        resource_name = os.path.basename(destination)
+        # The local path is decided according to whether we are in upgrade
+        local_resource_path = self._get_resource_file_path(service_name,
+                                                           resource_name,
+                                                           self.is_upgrade)
+
+        if not os.path.isfile(local_resource_path):
+            mkdir(os.path.dirname(local_resource_path))
+            if user_resource:
```
adamlavie: render should propagate to download_user_resource

components/utils.py
```diff
@@ -520,3 +620,361 @@ def untar(source, destination='/tmp', strip=1):
...
+                install_resource = self._get_resource_file_path(
+                        service_name, resource_name, False)
+                copy(existing_resource_path, install_resource)
+                # copy it to upgrade-resource_name
+                copy(existing_resource_path, dest)
+            else:
+                ctx.logger.info('User resource {0} not found on {1}'
+                                .format(resource_name, path=dest))
+
+        if not os.path.isfile(dest):
```
adamlavie: render fix here

components/utils.py
```diff
@@ -520,3 +620,361 @@ def untar(source, destination='/tmp', strip=1):
...
+                        .format(resource_name, dest))
+        # rendering will be possible only for runtime properties
+        tmp_file = ctx.download_resource(source)
+        move(tmp_file, dest)
+
+    def _download_resource_and_render(self, source, dest, service_name,
+                                      load_ctx):
+        if load_ctx:
+            params = self._load_node_props(service_name)
+        else:
```
adamlavie: If non simply don't pass params

components/utils.py
```diff
@@ -520,3 +620,361 @@ def untar(source, destination='/tmp', strip=1):
...
+    def _download_user_resource(self, source, dest, resource_name,
+                                service_name, load_ctx=True):
+        if self.is_upgrade:
+            install_props = self.get_install_resources_json(service_name)
+            existing_resource_path = install_props.get(resource_name, '')
+            if os.path.isfile(existing_resource_path):
+                ctx.logger.info('Using existing resource for {0}'
+                                .format(resource_name))
+                # update the resource file we hold that might have changed
+                install_resource = self._get_resource_file_path(
```
adamlavie: make false as named arg

components/utils.py
```diff
@@ -520,3 +620,361 @@ def untar(source, destination='/tmp', strip=1):
...
+        if upgrade:
+            upgrade_resource_name = self.UPGRADE_RESOURCE_PREFIX + \
+                                    resource_name
+
+            dest_file_path = os.path.join(base_service_res_dir,
+                                          upgrade_resource_name)
+        else:
+            dest_file_path = os.path.join(base_service_res_dir, resource_name)
+        return dest_file_path
+
```
adamlavie: should be private

components/utils.py
```diff
@@ -520,3 +620,361 @@ def untar(source, destination='/tmp', strip=1):
...
+    def get_upgrade_resources_json(self, service_name):
+        return self._get_resources_json(service_name, True)
+
```
adamlavie: shyould be private

components/utils.py
```diff
@@ -520,3 +622,356 @@ def untar(source, destination='/tmp', strip=1):
...
+                                             local_resource_path,
+                                             resource_name,
+                                             service_name,
+                                             render=render,
+                                             load_ctx=load_ctx)
+            elif source_resource:
+                self._download_source_resource(source,
+                                               local_resource_path)
+            elif render:
+                self._download_resource_and_render(source,
```
nir0s: since render passes to `download_user_resource`... is there a reason to also have it here?

PR #281, opened 2016-04-25
adding-custom-es-scripts-via-create

max-orlov:


components/elasticsearch/scripts/create.py
```diff
@@ -186,6 +186,13 @@ def _install_elasticsearch():
     utils.chown('elasticsearch', 'elasticsearch',
                 os.path.join(es_conf_path, 'logging.yml'))
 
+    ctx.logger.info("Creating Elasticsearch scripts folder and "
```
nir0s: single quotes please. Also, per our convention in the manager blueprints, please add `...` at the end.

PR #287, opened 2016-05-05
CFY-5253 Openstack manager blueprint should create a subnet of class B

pawelcodilime:


openstack-manager-blueprint.yaml
```diff
@@ -687,7 +687,7 @@ node_templates:
       resource_id: { get_input: management_subnet_name }
       subnet:
         ip_version: 4
-        cidr: 10.67.79.0/24
+        cidr: 10.67.0.0/16
```
ran-z: actually, while this will obviously work, it's custom that class A addresses only go up to 127.X.X.X (and the default is indeed usually 10.X.X.X), class B is 128-191, and class C is 192+
so possibly should change this to 128.X.0.0/16?
what do you think?
(as you can see, before, it was class C and also started with 10, so..)

pawelcodilime: Well, it seems kind of reasonable, but aren't addresses 128.x.x.x from the public range?

PR #291, opened 2016-05-09
CFY-5225 create snapshot and dump snapshot and provider_context

adamlavie:


components/elasticsearch/scripts/create.py
```diff
@@ -17,15 +17,18 @@
 CONFIG_PATH = "components/elasticsearch/config"
 ES_SERVICE_NAME = 'elasticsearch'
 
-ctx_properties = utils.ctx_factory.create('elasticsearch')
+UPGRADE_DUMP_PATH = '/tmp/es_upgrade_dump/'
```
nir0s: these are internal, why do we need them as global variables?

components/elasticsearch/scripts/create.py
```diff
@@ -262,14 +267,96 @@ def main():
...
+    es_props = utils.ctx_factory.get(ES_SERVICE_NAME, upgrade_props=False)
+    if es_props['es_endpoint_ip']:
+        es_endpoint = es_props['es_endpoint_ip']
+    else:
+        es_endpoint = ctx.instance.host_ip
+    return es_endpoint
+
+
+def dump_upgrade_data():
+
```
nir0s: I don't quite understand. aren't these exactly the things that create the parity?

components/restservice/scripts/create.py
```diff
@@ -60,8 +60,9 @@ def install_optional(rest_venv):
         ctx.logger.info('Installing REST Service...')
         utils.install_python_package('/tmp/rest-service', rest_venv)
         ctx.logger.info('Deploying Required Manager Resources...')
-        utils.move(
+        utils.copy(
```
nir0s: ?

components/riemann/scripts/create.py
```diff
@@ -94,3 +94,5 @@ def install_riemann():
     utils.clean_var_log_dir(RIEMANN_SERVICE_NAME)
 
 install_riemann()
+if utils.is_upgrade:
```
nir0s: why not create the dummy node for nirb to use?

components/utils.py
```diff
@@ -14,6 +14,9 @@
 import subprocess
 from functools import wraps
 import json
+import urllib2
```
nir0s: the imports are organized according to length. please keep it that way.

components/utils.py
```diff
@@ -135,11 +141,14 @@ def error_exit(message):
     sys.exit(1)
 
 
-def mkdir(dir):
+def mkdir(dir, use_sudo=True):
     if os.path.isdir(dir):
         return
     ctx.logger.info('Creating Directory: {0}'.format(dir))
-    sudo(['mkdir', '-p', dir])
+    if use_sudo:
```
nir0s: maybe 

```python
cmd = ['mkdir', '-p', directory]

if use_sudo:
    sudo(cmd)
else:
    run(cmd)
```

components/utils.py
```diff
@@ -705,65 +714,75 @@ class CtxPropertyFactory(object):
...
-            # write properties to a local backup path
-            self._write_props_to_file(ctx_props, service_name)
+        self._write_props_to_file(ctx_props, service_name)
 
         return ctx_props
 
-    def _write_props_to_file(self, ctx_props, service_name):
+    def get(self, service_name, upgrade_props=True):
+        """
+        get node properties by service name.
```
nir0s: Please keep the format of the docstrings

capital letter, first line in the same line where the triple double-quotes are, line break between the explanation and the params..

components/utils.py
```diff
@@ -705,65 +714,75 @@ class CtxPropertyFactory(object):
...
+            props = self._get_install_properties(service_name)
+        return props
 
+    def _write_props_to_file(self, ctx_props, service_name):
         dest_file_path = self._get_props_file_path(service_name)
-        ctx.logger.info('Saving {service} input configuration to {path}'
-                        .format(service=service_name,
-                                path=dest_file_path))
-        write_to_json_file(ctx_props, dest_file_path)
+        if not os.path.exists(dest_file_path):
```
nir0s: exists looks for either a file or a directory.

components/utils.py
```diff
@@ -705,65 +714,75 @@ class CtxPropertyFactory(object):
...
+            ctx.logger.info('Saving {service} input configuration to {path}'
```
nir0s: please be consistent with the formatting as we talked about

components/utils.py
```diff
@@ -1004,20 +995,137 @@ def _get_rollback_resources_dir(self, service_name):
...
+        reqstring = url + (' ' + data if data else '')
+        ctx.logger.error('Failed to {0} {1} (reason: {2})'.format(
+                method, reqstring, e.reason))
+
+
+def _wait_for_execution(execution_id, headers):
+    poll_interval = 2
+    url = 'http://localhost/api/v2/executions?' \
+          '_include_system_workflows=true&id={0}'.format(execution_id)
+    while True:
```
nir0s: this can be stuck forever

components/utils.py
```diff
@@ -1004,20 +995,137 @@ def _get_rollback_resources_dir(self, service_name):
...
+        res = http_request(url, method='GET', headers=headers)
+        if res.code != 200:
+            raise RuntimeError('Failed listing existing executions.')
```
nir0s: what about returning the response itself in the exception..?

components/utils.py
```diff
@@ -1004,20 +995,137 @@ def _get_rollback_resources_dir(self, service_name):
...
+        return urllib2.urlopen(request)
+    except urllib2.URLError as e:
+        reqstring = url + (' ' + data if data else '')
+        ctx.logger.error('Failed to {0} {1} (reason: {2})'.format(
+                method, reqstring, e.reason))
+
+
+def _wait_for_execution(execution_id, headers):
+    poll_interval = 2
+    url = 'http://localhost/api/v2/executions?' \
```
nir0s: shouldn't this be the internal rest port?

components/utils.py
```diff
@@ -1004,20 +995,137 @@ def _get_rollback_resources_dir(self, service_name):
...
+    data = json.dumps({'include_metrics': 'true',
+                       'include_credentials': 'true'})
+    headers = _create_maintenance_headers(upgrade_props=False)
+    req_headers = headers.copy()
+    req_headers.update({'Content-Type': 'application/json'})
+    ctx.logger.info('Creating snapshot with ID {0}'
+                    .format(snapshot_id))
+    res = http_request(url, data=data, method='PUT', headers=req_headers)
+    if res.code != 201:
+        raise RuntimeError('Failed creating snapshot {0}'
```
nir0s: same, should return the response in the exception

components/utils.py
adamlavie: what do you suggest?

components/elasticsearch/scripts/create.py
adamlavie: In this part we dump the provider context and snapshots to a file and restore all the data after es has been upgraded.

nir0s: didn't we say we're going to add `provider_context` backup to the snapshots mechanism?

components/utils.py
nir0s: a timeout... or retry count..

components/elasticsearch/scripts/create.py
adamlavie: These are script constants. I can make them start with '_' if that's what you mean

components/elasticsearch/scripts/create.py
adamlavie: That did come up, but was not decided. This solution would require some changes to the snapshot API and after discussing with Dan and considering the time constraints, we thought this should be fine. Is there a specific reason why you think we should use the snapshots API.

components/riemann/scripts/create.py
adamlavie: Nirb actually has a branch in progress with this node

components/utils.py
adamlavie: So you think we should put a timeout on the snapshot restore process? This process duration really depends on how many deployments a user may have. What timeout would you give it?

components/restservice/scripts/create.py
adamlavie: move is not idempotent. We should consider changing the move impl since it will throw an error if files are already in place as in the above example.

components/elasticsearch/scripts/create.py
nir0s: nm

components/elasticsearch/scripts/create.py
nir0s: nm. we already talked about it several times.

components/elasticsearch/scripts/create.py
```diff
@@ -262,14 +267,96 @@ def main():
...
+                                                                port)
+    types = ['provider_context', 'snapshot']
+    ctx.logger.info('Dumping upgrade data: {0}'.format(types))
+    type_values = []
+    for _type in types:
+        res = http_request('{0}/_search?q=_type:{1}&size=10000'
+                           .format(storage_endpoint, _type),
+                           method='GET')
+        if not res.code == 200:
+            raise RuntimeError('Failed fetching type {0} from cloudify_storage'
```
nir0s: ctx.abort_operation(message).
Any raise should be replaced with it to terminate the entire operation's execution

components/restservice/scripts/create.py
nir0s: then let's do that instead.

components/utils.py
```diff
@@ -1004,20 +993,165 @@ def _get_rollback_resources_dir(self, service_name):
...
+    return headers
+
+
+def _get_auth_headers(upgrade_props):
+    headers = {}
+    config = ctx_factory.get('manager-config', upgrade_props=upgrade_props)
+    security = config['security']
+    security_enabled = security['enabled']
+    if security_enabled:
+        username = security.get('admin_username')
```
nir0s: can this change?

components/utils.py
```diff
@@ -1004,20 +993,165 @@ def _get_rollback_resources_dir(self, service_name):
...
+        password = security.get('admin_password')
+        headers.update({'Authorization':
+                        'Basic ' + base64.b64encode('{0}:{1}'.format(
+                                    username, password))})
+    return headers
+
+
+def create_upgrade_snapshot():
+    snapshot_id = _generate_upgrade_snapshot_id()
+    url = 'http://localhost/api/v2/snapshots/{0}'.format(snapshot_id)
```
nir0s: If i'm not mistaken, we already have v2.1, right?

components/utils.py
```diff
@@ -1004,20 +993,165 @@ def _get_rollback_resources_dir(self, service_name):
...
+    req_headers = headers.copy()
+    req_headers.update({'Content-Type': 'application/json'})
+    ctx.logger.info('Creating snapshot with ID {0}'
+                    .format(snapshot_id))
+    res = http_request(url, data=data, method='PUT', headers=req_headers)
+    if res.code != 201:
+        err = 'Failed creating snapshot {0}. Message: {1}'\
+              .format(snapshot_id, res.readlines())
+        ctx.logger.error(err)
+        raise RuntimeError(err)
```
nir0s: ctx.abort...

components/utils.py
```diff
@@ -1004,20 +993,165 @@ def _get_rollback_resources_dir(self, service_name):
...
+    headers = _create_maintenance_headers(upgrade_props=True)
+    req_headers = headers.copy()
+    req_headers.update({'Content-Type': 'application/json'})
+    ctx.logger.info('Restoring snapshot with ID {0}'.format(snapshot_id))
+    res = http_request(url, data=data, method='POST', headers=req_headers)
+    if res.code != 200:
+        err = 'Failed restoring snapshot {0}. Message: {1}' \
+              .format(snapshot_id, res.readlines())
+        ctx.logger.error(err)
+        raise RuntimeError(err)
```
nir0s: ctx.abort...

components/utils.py
```diff
@@ -1004,20 +993,165 @@ def _get_rollback_resources_dir(self, service_name):
...
+
+def _generate_upgrade_snapshot_id():
+    url = 'http://localhost/api/v1/version'
+    auth_headers = _get_auth_headers(upgrade_props=False)
+    res = http_request(url, method='GET', headers=auth_headers)
+    if res.code != 200:
+        err = 'Failed extracting current manager version. Message: {0}'\
+              .format(res.readlines())
+        ctx.logger.error(err)
+        raise RuntimeError(err)
```
nir0s: ctx.abort...

tests/unit-tests/test_upgrade.py
```diff
@@ -141,60 +140,55 @@ def assert_rpm_url_overridden(self, ctx_properties):
...
+        resource_json = utils.resource_factory._get_resources_json(
+                TEST_SERVICE_NAME)
 
         # assert resource json contains mapping to the new resource dest
         self.assertEqual(resource_json.get(TEST_RESOURCE_NAME),
                          resource_file_dest)
         resource_local_path = os.path.join(
-            resource_factory._get_resources_dir(TEST_SERVICE_NAME),
-            TEST_RESOURCE_NAME)
+                utils.resource_factory._get_resources_dir(TEST_SERVICE_NAME),
```
nir0s: all private functions used outside the factory class shouldn't be private

components/utils.py
adamlavie: Generally, it is supported yes

PR #295, opened 2016-05-11
CFY-5255 Perform basic sanity checks after starting each node

tehasdf:
Perform basic sanity checks in each node's start script, verifying
that the services started correctly.

### DIFF
--- a/components/elasticsearch/scripts/start.py
+++ b/components/elasticsearch/scripts/start.py
@@ -1,5 +1,8 @@
 #!/usr/bin/env python
 
+import json
+import urllib2
+import urlparse
 from os.path import join, dirname
 
 from cloudify import ctx
@@ -8,11 +11,49 @@ ctx.download_resource(
     join('components', 'utils.py'),
     join(dirname(__file__), 'utils.py'))
 import utils  # NOQA
-ES_SERVICE_NAME = 'elasticsearch'
 
+ES_SERVICE_NAME = 'elasticsearch'
 ctx_properties = utils.ctx_factory.get(ES_SERVICE_NAME)
-ES_ENDPOINT_IP = ctx_properties.get('es_endpoint_ip')
+ES_ENDPOINT_IP = ctx_properties['es_endpoint_ip']
+ES_ENDPOINT_PORT = ctx_properties['es_endpoint_port']
+
+
+def _examine_status_response(response):
+    """Check if the status response from elasticsearch is correct.
+
+    A correct response has a status of 200, and also returns a JSON object
+    with {'status': 200}.
+    Elasticsearch doesn't start up immediately, so this retries a few times.
+    """
+    if response.code != 200:
+        return False
+
+    parsed_response = json.load(response)
+
+    return parsed_response['status'] == 200
+
+
+def check_index_exists(url, index_name='cloudify_storage'):
+    """Check that the cloudify_storage ES index exists."""
+    index_url = urlparse.urljoin(url, index_name)
+    try:
+        return urllib2.urlopen(index_url)
+    except urllib2.URLError as e:
+        if e.code == 404:
+            ctx.abort_operation('The index {0} does not exist in ES'.format(
+                index_name))
+        else:
+            ctx.abort_operation('Invalid ES response: {0}'.format(e))
+
 
 if not ES_ENDPOINT_IP:
     ctx.logger.info('Starting Elasticsearch Service...')
     utils.start_service(ES_SERVICE_NAME, append_prefix=False)
+    ES_ENDPOINT_IP = '127.0.0.1'
+    utils.systemd.verify_alive(ES_SERVICE_NAME, append_prefix=False)
+
+elasticsearch_url = 'http://{0}:{1}/'.format(ES_ENDPOINT_IP,
+                                             ES_ENDPOINT_PORT)
+utils.verify_service_http(ES_SERVICE_NAME, elasticsearch_url,
+                          _examine_status_response)
+check_index_exists(elasticsearch_url)
--- a/components/influxdb/scripts/start.py
+++ b/components/influxdb/scripts/start.py
@@ -14,9 +14,30 @@ INFLUX_SERVICE_NAME = 'influxdb'
 
 ctx_properties = utils.ctx_factory.get(INFLUX_SERVICE_NAME)
 
+INFLUXDB_ENDPOINT_IP = ctx_properties['influxdb_endpoint_ip']
+INFLUXDB_ENDPOINT_PORT = 8086
 
-INFLUXDB_ENDPOINT_IP = ctx_properties.get('influxdb_endpoint_ip')
+
+def check_influxdb_response(response):
+    """Check if the response from influxdb is correct.
+
+    InfluxDB normally responds with a 404 on GET to /, but also allow other
+    non-server-error response codes to allow for that behaviour to change.
+    """
+    return response.code < 500
+
 
 if not INFLUXDB_ENDPOINT_IP:
     ctx.logger.info('Starting InfluxDB Service...')
     utils.start_service(INFLUX_SERVICE_NAME)
+
+    INFLUXDB_ENDPOINT_IP = '127.0.0.1'
+
+    utils.systemd.verify_alive(INFLUX_SERVICE_NAME)
+
+influxdb_url = 'http://{0}:{1}'.format(
+    INFLUXDB_ENDPOINT_IP, INFLUXDB_ENDPOINT_PORT)
+
+
+utils.verify_service_http(INFLUX_SERVICE_NAME, influxdb_url,
+                          check_influxdb_response)
--- a/components/logstash/scripts/start.py
+++ b/components/logstash/scripts/start.py
@@ -11,6 +11,7 @@ import utils  # NOQA
 
 LOGSTASH_SERVICE_NAME = 'logstash'
 
-
 ctx.logger.info('Starting Logstash Service...')
 utils.start_service(LOGSTASH_SERVICE_NAME, append_prefix=False)
+
+utils.systemd.verify_alive(LOGSTASH_SERVICE_NAME, append_prefix=False)
--- a/components/mgmtworker/scripts/start.py
+++ b/components/mgmtworker/scripts/start.py
@@ -10,7 +10,31 @@ ctx.download_resource(
 import utils  # NOQA
 
 MGMT_WORKER_SERVICE_NAME = 'mgmtworker'
+CELERY_PATH = '/opt/mgmtworker/env/bin/celery'  # also hardcoded in create
+
+
+@utils.retry(ValueError)
+def check_worker_running(amqp_url):
+    """Use `celery status` to check if the worker is running."""
+    result = utils.sudo([
+        CELERY_PATH,
+        '-b', celery_amqp_url,
+        '--app=cloudify_agent.app.app',
+        'status'
+    ], ignore_failures=True)
+    if result.returncode != 0:
+        raise ValueError('celery status: worker not running')
 
 
 ctx.logger.info('Starting Management Worker Service...')
 utils.start_service(MGMT_WORKER_SERVICE_NAME)
+
+utils.systemd.verify_alive(MGMT_WORKER_SERVICE_NAME)
+celery_amqp_url = ('amqp://{rabbitmq_username}:{rabbitmq_password}@'
+                   '{rabbitmq_endpoint_ip}:{broker_port}//').format(
+    **ctx.instance.runtime_properties)
+
+try:
+    check_worker_running(celery_amqp_url)
+except ValueError:
+    ctx.abort_operation('Celery worker failed to start')
--- a/components/nginx/scripts/start.py
+++ b/components/nginx/scripts/start.py
@@ -8,8 +8,24 @@ ctx.download_resource(
     join('components', 'utils.py'),
     join(dirname(__file__), 'utils.py'))
 import utils  # NOQA
+
 NGINX_SERVICE_NAME = 'nginx'
+
+
+def check_response(response):
+    """Check if the response looks like a correct REST service response.
+
+    We can get a 200, or a 401 in case auth is enabled. We don't expect a
+    502, though, as this would mean nginx isn't correctly proxying to
+    the REST service.
+    """
+    return response.code in {200, 401}
 
 
 ctx.logger.info('Starting Nginx Service...')
 utils.start_service(NGINX_SERVICE_NAME, append_prefix=False)
+utils.systemd.verify_alive(NGINX_SERVICE_NAME, append_prefix=False)
+
+nginx_url = '{0}://127.0.0.1/api/v2.1/blueprints'.format(
+    ctx.instance.runtime_properties['rest_protocol'])
+utils.verify_service_http(NGINX_SERVICE_NAME, nginx_url, check_response)
--- a/components/rabbitmq/scripts/start.py
+++ b/components/rabbitmq/scripts/start.py
@@ -12,11 +12,22 @@ ctx.download_resource(
 import utils  # NOQA
 
 RABBITMQ_SERVICE_NAME = 'rabbitmq'
-
-
 ctx_properties = utils.ctx_factory.get(RABBITMQ_SERVICE_NAME)
-
 rabbitmq_endpoint_ip = ctx_properties['rabbitmq_endpoint_ip']
+PORT = 5671 if ctx_properties['rabbitmq_ssl_enabled'] else 5672
+
+
+@utils.retry(ValueError)
+def check_rabbit_running():
+    """Use rabbitmqctl status to check if RabbitMQ is working.
+
+    Sometimes rabbit takes a while to start, so this is retried several times.
+    Note that this is currently impossible to do on a remote host, so
+    this check only runs when rabbitmq is installed locally.
+    """
+    result = utils.sudo(['rabbitmqctl', 'status'], ignore_failures=True)
+    if result.returncode != 0:
+        raise ValueError('rabbitmqctl status: rabbitmq not running')
 
 
 def set_rabbitmq_policy(name, expression, policy):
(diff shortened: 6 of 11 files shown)


components/elasticsearch/scripts/start.py
```diff
@@ -9,11 +12,52 @@
...
 import utils  # NOQA
 ES_SERVICE_NAME = 'elasticsearch'
+RETRIES = 10
 
 ctx_properties = utils.ctx_factory.create(ES_SERVICE_NAME, write_to_file=False)
 ES_ENDPOINT_IP = ctx_properties['es_endpoint_ip']
+ES_ENDPOINT_PORT = ctx_properties['es_endpoint_port']
+
+
+def is_elasticsearch_responding(url):
```
tehasdf: generalize and factor out to utils? `check_http(url, pred)`

components/elasticsearch/scripts/start.py
```diff
@@ -9,11 +12,52 @@
...
                                                append_prefix=False)
+    ES_ENDPOINT_IP = '127.0.0.1'
+    if utils.systemd.is_alive(ES_SERVICE_NAME):
+        ctx.logger.info('Elasticsearch service is running')
+    else:
+        ctx.logger.error('Elasticsearch service is not running')
+
+elasticsearch_url = 'http://{}:{}/'.format(ES_ENDPOINT_IP,
+                                           ES_ENDPOINT_PORT)
+for retry in range(RETRIES):
```
tehasdf: factor out to utils? `retry(timeout)`

components/elasticsearch/scripts/start.py
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env python
 
+import json
```
nir0s: please arrange dependencies by length and any "from"s should be at the end

components/elasticsearch/scripts/start.py
nir0s: yes. definitely

nir0s:
I think there's a design issue here. We need to check not only that services are running and responding to http calls (etc..) but that they're also configured. Now, I understand that checking the entire configuration is impossible, but there are some basic checks we can do to verify that at least some configuration is there. For instance, checking that elasticsearch is alive is nice, but maybe you can check that the `cloudify_storage` index exists and that it contains `provider_context`. Since the REST service depends on elasticsearch, maybe you can check that it's able to make a simple call (to any endpoint) and not just that its port is opened.. .etc..

components/elasticsearch/scripts/start.py
nir0s: I'm pretty sure we have it in utils already

components/elasticsearch/scripts/start.py
```diff
@@ -9,11 +12,52 @@
...
+    # ES takes a while to start
+    if is_elasticsearch_responding(elasticsearch_url):
+        ctx.logger.info('ES responding')
+        break
+    else:
+        ctx.logger.info('ES not responding (retrying {}... {})'.format(
+            elasticsearch_url, retry + 1))
+        time.sleep(3)
+else:
+    ctx.logger.error('ES didnt respond in {} tries'.format(RETRIES))
```
nir0s: {} is limited to Python 2.7+. While we only support RHEL/Centos 7 for the Manager which already come with Python 2.7, please use {0}... so that we can still support 2.6.

components/influxdb/scripts/start.py
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 
 from os.path import join, dirname
+import time
```
nir0s: again, order

components/influxdb/scripts/start.py
```diff
@@ -17,7 +19,47 @@
 
 
 INFLUXDB_ENDPOINT_IP = ctx_properties['influxdb_endpoint_ip']
+INFLUXDB_ENDPOINT_PORT = 8086
+RETRIES = 10
+
+
+def is_influxdb_responding(url):
+    """Check if url responds with a normal influxdb response."""
```
nir0s: utils

components/elasticsearch/scripts/start.py
```diff
@@ -9,11 +12,52 @@
...
 
 if not ES_ENDPOINT_IP:
     ctx.logger.info('Starting Elasticsearch Service...')
     utils.start_service_and_archive_properties(ES_SERVICE_NAME,
                                                append_prefix=False)
+    ES_ENDPOINT_IP = '127.0.0.1'
+    if utils.systemd.is_alive(ES_SERVICE_NAME):
+        ctx.logger.info('Elasticsearch service is running')
+    else:
+        ctx.logger.error('Elasticsearch service is not running')
```
nir0s: if it's not running... shouldn't we do `ctx.abort_operation` here?

components/elasticsearch/scripts/start.py
nir0s: also here.. shouldn't this abort the operation?

components/influxdb/scripts/start.py
```diff
@@ -17,7 +19,47 @@
...
+        return False
+    else:
+        return True
+
 
 if not INFLUXDB_ENDPOINT_IP:
     ctx.logger.info('Starting InfluxDB Service...')
     utils.start_service_and_archive_properties(INFLUX_SERVICE_NAME)
+
+    INFLUXDB_ENDPOINT_IP = '127.0.0.1'
```
nir0s: this entire section below looks like a duplication of the elasticsearch one. Why not create a utils class which is instantiated for the relevant service and performs the relevant checks?

components/logstash/scripts/start.py
```diff
@@ -15,3 +15,8 @@
 ctx.logger.info('Starting Logstash Service...')
 utils.start_service_and_archive_properties(LOGSTASH_SERVICE_NAME,
                                            append_prefix=False)
+
+if utils.systemd.is_alive(LOGSTASH_SERVICE_NAME, append_prefix=False):
```
nir0s: the logging for each `is_alive` check should probably move to the `is_alive` method as it's repeated everywhere.

components/mgmtworker/scripts/start.py
```diff
@@ -10,7 +11,44 @@
...
+    return resp.returncode == 0
 
 
 ctx.logger.info('Starting Management Worker Service...')
 utils.start_service_and_archive_properties(MGMT_WORKER_SERVICE_NAME)
+
+if utils.systemd.is_alive(MGMT_WORKER_SERVICE_NAME):
+    ctx.logger.info('Management Worker Service is running')
+else:
+    ctx.logger.error('Management Worker Service is not running')
```
nir0s: again, if a service is not running, the operation should be aborted. There's no situation in which the service is not running (after retries maybe) and it's ok.

components/rabbitmq/scripts/start.py
```diff
@@ -12,7 +13,7 @@
 import utils  # NOQA
 
 RABBITMQ_SERVICE_NAME = 'rabbitmq'
-
+PORT = 5672
```
nir0s: when rabbit is working with SSL, the port is not 5672. See `create.py`

components/rabbitmq/scripts/start.py
```diff
@@ -94,3 +95,25 @@ def set_rabbitmq_policy(name, expression, policy):
...
+
+    if utils.systemd.is_alive(RABBITMQ_SERVICE_NAME):
+        ctx.logger.info('RabbitMQ service is running')
+    else:
+        ctx.logger.info('RabbitMQ service is not running')
+
+
+sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+try:
```
nir0s: another cross-script comment: please put everything in functions so that they can be easily unit tested

components/influxdb/scripts/start.py
```diff
@@ -10,14 +10,32 @@
...
+
+    utils.systemd.verify_alive(INFLUX_SERVICE_NAME)
+
+influxdb_url = 'http://{}:{}'.format(
```
nir0s: {0} formatting please

components/mgmtworker/scripts/start.py
```diff
@@ -10,7 +11,30 @@
...
+CELERY_PATH = '/opt/mgmtworker/env/bin/celery'  # also hardcoded in create
+
+ctx_properties = utils.ctx_factory.get_install_properties(
+    MGMT_WORKER_SERVICE_NAME)
+
+
+@utils.retry(subprocess.CalledProcessError)
+def check_worker_running(amqp_url):
+    """Use `celery status` to check if the worker is running."""
+    return utils.sudo([
```
nir0s: you seem to be returning here even though you're not doing anything with the return value..

components/restservice/scripts/start.py
```diff
@@ -10,7 +12,39 @@
...
+
+
+def verify_restservice(url, credentials=None):
+    """To verify that the REST service is working, GET the blueprints list.
+
+    There's nothing special about the blueprints endpoint, it's simply one
+    that also requires the storage backend to be up, so if it works, there's
+    a good chance everything is configured correctly.
+    """
+    blueprints_url = urlparse.urljoin(url, 'api/v2/blueprints')
```
nir0s: I believe there's already a v2.1. need to verify

components/restservice/scripts/start.py
```diff
@@ -10,7 +12,39 @@
...
+    req = urllib2.Request(blueprints_url)
+
+    if credentials is not None:
```
nir0s: if credentials?

components/restservice/scripts/start.py
```diff
@@ -10,7 +12,39 @@
...
+                            .format(e))
+        raise
 
 
 ctx.logger.info('Starting Cloudify REST Service...')
 utils.start_service_and_archive_properties(REST_SERVICE_NAME)
+
+utils.systemd.verify_alive(REST_SERVICE_NAME)
+
+restservice_url = 'http://{}:{}'.format('127.0.0.1', 8100)
```
nir0s: formatting

components/riemann/scripts/start.py
```diff
@@ -14,3 +14,8 @@
 
 ctx.logger.info('Starting Riemann Service...')
 utils.start_service_and_archive_properties(RIEMANN_SERVICE_NAME)
+
+if utils.systemd.is_alive(RIEMANN_SERVICE_NAME):
```
nir0s: replace with `verify_alive`

components/influxdb/scripts/start.py
```diff
@@ -14,9 +14,30 @@
...
+    non-server-error response codes to allow for that behaviour to change.
+    """
+    return response.code < 500
 
-INFLUXDB_ENDPOINT_IP = ctx_properties.get('influxdb_endpoint_ip')
 
 if not INFLUXDB_ENDPOINT_IP:
     ctx.logger.info('Starting InfluxDB Service...')
-    utils.start_service(INFLUX_SERVICE_NAME)
+    utils.start_service_and_archive_properties(INFLUX_SERVICE_NAME)
```
nir0s: I believe this is now `start_service`.

components/java/scripts/start.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python
```
nir0s: can we call this `validate.py`?

Ideally, I would like the validations to take place AFTER `start` in each node, but since there's no way to add a validate lifecycle operation, `start` is a good place. In the case of `java` and `python`, there IS no start, so creating a `validate.py` script makes more sense.

components/java/scripts/start.py
```diff
@@ -0,0 +1,15 @@
...
+
+ctx.download_resource(
+    join('components', 'utils.py'),
+    join(dirname(__file__), 'utils.py'))
+import utils  # NOQA
+
+
+java_result = utils.sudo(['java', '-version'], ignore_failures=True)
+if java_result.returncode != 0:
+    ctx.abort_operatoin('Java runtime error: java was not installed')
```
nir0s: operation

components/elasticsearch/scripts/start.py
```diff
@@ -8,11 +11,50 @@
...
+    """
+    if response.code != 200:
+        return False
+
+    parsed_response = json.load(response)
+
+    return parsed_response['status'] == 200
+
+
+def check_index_exists(url, index_name='cloudify_storage'):
```
nir0s: I'm wondering - do you think there's a reason to use the retry decorator here? Obviously, you first check that an http response is available, so there's no reason why this wouldn't work the first time. Thing is, I'm not sure we can account for any potential problem here. What do you think?

tehasdf: Well we retry the first http response check (in `verify_index_http`), so by the time this runs, everything is up and running.

Only thing I see that we could possibly guard against by using a retry here, is a network failure (with an external ES)... but if the network is that bad, can we really continue?

nir0s: yeah.. good point.

components/nginx/scripts/start.py
```diff
@@ -8,8 +8,12 @@
...
 import utils  # NOQA
-NGINX_SERVICE_NAME = 'nginx'
 
+NGINX_SERVICE_NAME = 'nginx'
 
 ctx.logger.info('Starting Nginx Service...')
 utils.start_service(NGINX_SERVICE_NAME, append_prefix=False)
+utils.systemd.verify_alive(NGINX_SERVICE_NAME, append_prefix=False)
+
+nginx_url = 'http://127.0.0.1'
```
nir0s: Let's try and understand if we're using SSL here or not. If https is enabled, we should probably check 443 instead.

components/rabbitmq/scripts/start.py
```diff
@@ -92,3 +100,9 @@ def set_rabbitmq_policy(name, expression, policy):
 
     # rabbitmq restart exits with 143 status code that is valid in this case.
     utils.start_service(RABBITMQ_SERVICE_NAME, ignore_restart_fail=True)
+    rabbitmq_endpoint_ip = '127.0.0.1'
+
+    utils.systemd.verify_alive(RABBITMQ_SERVICE_NAME)
+
+utils.verify_port_open(RABBITMQ_SERVICE_NAME, PORT, host=rabbitmq_endpoint_ip)
+check_rabbit_running()
```
nir0s: um, this is true only if the endpoint is localhost.

components/restservice/scripts/start.py
```diff
@@ -10,7 +12,41 @@
...
+
+
+def verify_restservice(url, username='', password=''):
+    """To verify that the REST service is working, GET the blueprints list.
+
+    There's nothing special about the blueprints endpoint, it's simply one
+    that also requires the storage backend to be up, so if it works, there's
+    a good chance everything is configured correctly.
+    """
+    blueprints_url = urlparse.urljoin(url, 'api/v2.1/blueprints')
```
nir0s: FYI, https://cloudifysource.atlassian.net/browse/CFY-5316

components/restservice/scripts/start.py
```diff
@@ -10,7 +12,41 @@
...
+    if username or password:
+        auth_header = utils.basic_auth_header(username, password)
+        req.add_header('Authorization', auth_header)
+
+    try:
+        return urllib2.urlopen(req)
+    except urllib2.URLError as e:
+        ctx.abort_operation('REST service returned an invalid response: {0}'
+                            .format(e))
+        raise
```
nir0s: no need to raise after an `abort_operation`..

components/utils.py
```diff
@@ -287,6 +287,12 @@ def copy_notice(service):
     copy(resource_file, dest)
 
 
+def is_port_open(port, host='localhost'):
```
nir0s: utils.wait_for_port

components/nginx/scripts/start.py
```diff
@@ -8,8 +8,25 @@
...
+    the REST service.
+    """
+    return response.code in {200, 401}
 
 
 ctx.logger.info('Starting Nginx Service...')
 utils.start_service(NGINX_SERVICE_NAME, append_prefix=False)
+utils.systemd.verify_alive(NGINX_SERVICE_NAME, append_prefix=False)
+
+nginx_url = '{0}://127.0.0.1/api/v2.1/blueprints'.format(
```
tehasdf: Similar to how we hit the endpoint in restservice start script (to check that the restservice can access ES correctly), here we can try to access restservice via nginx, to check if it's proxying correctly? Do we want this check to be more strict, and do we want it at all?

components/mgmtworker/scripts/start.py
```diff
@@ -10,7 +11,30 @@
...
 
 
 ctx.logger.info('Starting Management Worker Service...')
 utils.start_service(MGMT_WORKER_SERVICE_NAME)
+
+utils.systemd.verify_alive(MGMT_WORKER_SERVICE_NAME)
+
+celery_amqp_url = ('amqp://{rabbitmq_username}:{rabbitmq_password}@'
+                   '{rabbitmq_endpoint_ip}:{port}//').format(
+    port=5672, **ctx_properties)
```
nir0s: um.. SSL?

components/nginx/scripts/start.py
nir0s: it makes sense to test that it's proxying correctly.

types/manager-types.yaml
```diff
@@ -291,6 +303,14 @@ node_types:
               default: components/rabbitmq/scripts/stop.py
             fabric_env: *simple_fabric_env
 
+      cloudify.interfaces.sanitycheck:
+        postinstall:
+          implementation: fabric.fabric_plugin.tasks.run_script
+          inputs:
+            script_path:
+              default: components/checks/rabbitmq_check.py
```
nir0s: ?

tehasdf: oh whoops, sorry, i broke a merge and committed this stuff that i used for debugging some time ago :X

components/mgmtworker/scripts/start.py
```diff
@@ -10,7 +10,31 @@
 import utils  # NOQA
 
 MGMT_WORKER_SERVICE_NAME = 'mgmtworker'
+CELERY_PATH = '/opt/mgmtworker/env/bin/celery'  # also hardcoded in create
+
+
+@utils.retry(ValueError)
```
tehasdf: `ValueError` isn't the ideal choice, but it's the closest of built-in exceptions IMO - the only real alternative being `RuntimeError` which is too general i think. And the exception doesn't leave this module anyway, so it's not a big problem.

components/rabbitmq/scripts/start.py
```diff
@@ -12,11 +12,22 @@
...
+@utils.retry(ValueError)
+def check_rabbit_running():
+    """Use rabbitmqctl status to check if RabbitMQ is working.
 
-rabbitmq_endpoint_ip = ctx_properties['rabbitmq_endpoint_ip']
+    Sometimes rabbit takes a while to start, so this is retried several times.
+    Note that this is currently impossible to do on a remote host, so
+    this check only runs when rabbitmq is installed locally.
+    """
+    result = utils.sudo(['rabbitmqctl', 'status'], ignore_failures=True)
```
tehasdf: in the future we might want to look into getting `rabbitmqctl status` working also remotely, which will allow us to use this check also for external rabbitmq, instead of the - less strict - `verify_port_open` check

PR #175, opened 2015-11-10
CFY-3691 - Make all internal REST calls secure

noakup:

PR #297, opened 2016-05-12
CFY-4953 Pre-upgrade validations

tehasdf:
Add validation scripts to run before a manager upgrade,
checking that the required services are running, that the
directory structure looks as expected, and that some
properties didn't change.

### DIFF
--- a/components/elasticsearch/scripts/creation_validation.py
+++ b/components/elasticsearch/scripts/creation_validation.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+
+import urllib2
+import urlparse
+from os.path import join, dirname
+
+from cloudify import ctx
+
+ctx.download_resource(
+    join('components', 'utils.py'),
+    join(dirname(__file__), 'utils.py'))
+import utils  # NOQA
+
+
+ES_SERVICE_NAME = 'elasticsearch'
+
+
+def verify_properties():
+    """Compare node properties and decide if upgrading is allowed."""
+    changed = utils.changed_upgrade_properties(ES_SERVICE_NAME)
+
+    if 'es_heap_size' in changed:
+        bootstrap_heap_size, upgrade_heap_size = changed['es_heap_size']
+        bootstrap_heap_size = utils.parse_jvm_heap_size(bootstrap_heap_size)
+        upgrade_heap_size = utils.parse_jvm_heap_size(upgrade_heap_size)
+
+        if upgrade_heap_size < bootstrap_heap_size:
+            ctx.abort_operation('Upgrading a Cloudify Manager with '
+                                'Elasticsearch Heap Size lower than what it '
+                                'was initially bootstrapped with is not '
+                                'allowed.')
+
+
+def verify_elasticsearch_running(url):
+    """Check that ES is running, and that it contains the provider context.
+
+    This is a sanity check that the manager we're upgrading was bootstrapped
+    correctly.
+    """
+    provider_context_url = urlparse.urljoin(url, 'cloudify_storage/'
+                                                 'provider_context/CONTEXT')
+    try:
+        urllib2.urlopen(provider_context_url)
+    except urllib2.URLError as e:
+        ctx.abort_operation('ES returned an error when getting the provider '
+                            'context: {0}'.format(e))
+        raise
+
+
+if utils.is_upgrade:
+    install_properties = utils.ctx_factory.get(ES_SERVICE_NAME)
+    ES_ENDPOINT_IP = install_properties['es_endpoint_ip']
+    ES_ENDPOINT_PORT = install_properties['es_endpoint_port']
+    if not ES_ENDPOINT_IP:
+        ES_ENDPOINT_IP = '127.0.0.1'
+        utils.systemd.verify_alive(ES_SERVICE_NAME, append_prefix=False)
+
+    elasticsearch_url = 'http://{0}:{1}'.format(ES_ENDPOINT_IP,
+                                                ES_ENDPOINT_PORT)
+
+    utils.validate_upgrade_directories(ES_SERVICE_NAME)
+    verify_elasticsearch_running(elasticsearch_url)
+    verify_properties()
--- a/components/manager/scripts/creation_validation.py
+++ b/components/manager/scripts/creation_validation.py
@@ -1,0 +1,19 @@
+#!/usr/bin/env python
+
+from os.path import join, dirname
+
+from cloudify import ctx
+
+ctx.download_resource(
+    join('components', 'utils.py'),
+    join(dirname(__file__), 'utils.py'))
+import utils  # NOQA
+
+IMMUTABLE_PROPERTIES = [
+    'ssl',
+    'security',
+    'ssh_user'
+]
+
+if utils.is_upgrade:
+    utils.verify_immutable_properties('manager-config', IMMUTABLE_PROPERTIES)
--- a/components/restservice/scripts/creation_validation.py
+++ b/components/restservice/scripts/creation_validation.py
@@ -1,0 +1,17 @@
+#!/usr/bin/env python
+
+from os.path import join, dirname
+
+from cloudify import ctx
+
+ctx.download_resource(
+    join('components', 'utils.py'),
+    join(dirname(__file__), 'utils.py'))
+import utils  # NOQA
+
+
+REST_SERVICE_NAME = 'restservice'
+
+if utils.is_upgrade:
+    utils.validate_upgrade_directories(REST_SERVICE_NAME)
+    utils.systemd.verify_alive(REST_SERVICE_NAME)
--- a/components/riemann/scripts/creation_validation.py
+++ b/components/riemann/scripts/creation_validation.py
@@ -1,0 +1,16 @@
+#!/usr/bin/env python
+
+from os.path import join, dirname
+
+from cloudify import ctx
+
+ctx.download_resource(
+    join('components', 'utils.py'),
+    join(dirname(__file__), 'utils.py'))
+import utils  # NOQA
+
+RIEMANN_SERVICE_NAME = 'riemann'
+
+if utils.is_upgrade:
+    utils.validate_upgrade_directories(RIEMANN_SERVICE_NAME)
+    utils.systemd.verify_alive(RIEMANN_SERVICE_NAME)
--- a/components/utils.py
+++ b/components/utils.py
@@ -776,7 +776,7 @@ class CtxPropertyFactory(object):
         if os.path.isfile(rollback_props_path):
             ctx.logger.info('Restoring service input properties for service '
                             '{0}'.format(service_name))
-            rollback_dir = self._get_rollback_properties_dir(service_name)
+            rollback_dir = self.get_rollback_properties_dir(service_name)
             install_dir = self._get_properties_dir(service_name)
             if os.path.isdir(install_dir):
                 remove(install_dir)
@@ -802,7 +802,7 @@ class CtxPropertyFactory(object):
         return dest_file_path
 
     def _get_rollback_props_file_path(self, service_name):
-        base_service_dir = self._get_rollback_properties_dir(service_name)
+        base_service_dir = self.get_rollback_properties_dir(service_name)
         dest_file_path = os.path.join(base_service_dir,
                                       self.PROPERTIES_FILE_NAME)
         return dest_file_path
@@ -830,7 +830,7 @@ class CtxPropertyFactory(object):
                             service_name,
                             self.NODE_PROPS_DIR_NAME)
 
-    def _get_rollback_properties_dir(self, service_name):
+    def get_rollback_properties_dir(self, service_name):
         return os.path.join(self.BASE_PROPERTIES_PATH,
                             service_name,
                             self.ROLLBACK_NODE_PROPS_DIR_NAME)
@@ -991,7 +991,7 @@ class BlueprintResourceFactory(object):
         return {'node': {'properties': node_props}}
 
     def _get_local_file_path(self, service_name, resource_name):
-        base_service_res_dir = self._get_resources_dir(service_name)
+        base_service_res_dir = self.get_resources_dir(service_name)
         dest_file_path = os.path.join(base_service_res_dir, resource_name)
         return dest_file_path
 
@@ -1009,7 +1009,7 @@ class BlueprintResourceFactory(object):
         write_to_json_file(resources_dict, resources_json)
 
     def _restore_resources(self, service_name):
-        rollback_dir = self._get_rollback_resources_dir(service_name)
+        rollback_dir = self.get_rollback_resources_dir(service_name)
         if not os.path.isdir(rollback_dir):
             # node resources have already been moved.
             return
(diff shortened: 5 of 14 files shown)


components/elasticsearch/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python
+
+import json
+
+from os.path import join, dirname
```
nir0s: order by length, from at the end. first, stdlib imports, then cloudify

components/elasticsearch/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,65 @@
...
+total_heap_bytes_used = sum(
+    node_data['jvm']['mem']['heap_used_in_bytes']
+    for node_data in es_jvm_stats['nodes'].values()
+)
+
+heap_used_in_mb = total_heap_bytes_used // (1024 ** 2)
+# TODO use the function from validate.py - factor out to utils
+es_heap_size = upgrade_properties['es_heap_size']
+
+if es_heap_size.endswith('g'):
```
nir0s: duplication from validate.py.

components/elasticsearch/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,65 @@
...
+    urllib2.urlopen('http://{}:9200/_nodes/stats/jvm'.format(
+        ES_ENDPOINT_IP)))
+
+total_heap_bytes_used = sum(
+    node_data['jvm']['mem']['heap_used_in_bytes']
+    for node_data in es_jvm_stats['nodes'].values()
+)
+
+heap_used_in_mb = total_heap_bytes_used // (1024 ** 2)
+# TODO use the function from validate.py - factor out to utils
```
nir0s: yup

components/elasticsearch/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,65 @@
...
+    multiplier = 1
+else:
+    raise ValueError('elasticsearch_heap_size input is invalid. '
+                     'The input size can be one of `Xm` or `Xg` formats. '
+                     'Provided: {0}'.format(es_heap_size))
+
+es_heap_size_in_mb = int(es_heap_size[:-1]) * multiplier
+
+if heap_used_in_mb > es_heap_size_in_mb:
+    raise RuntimeError('New Elasticsearch heap size ({}) is not enough for '
```
nir0s: this statement is not true. it's not that it's not enough, it's that we assume that if a user chose to bootstrap with, say 4GB of heap size, it means that upgrading using less might break things. It should state something like:

"Upgrading a Cloudify Manager with Elasticsearch Heap Size lower than what it was initially bootstrapped with is not allowed."

components/influxdb/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,18 @@
...
+    join(dirname(__file__), 'utils.py'))
+
+import utils  # NOQA
+
+INFLUX_SERVICE_NAME = 'influxdb'
+
+utils.upgrade_validation_directories(INFLUX_SERVICE_NAME)
+
+if not utils.systemd.is_alive(INFLUX_SERVICE_NAME):
+    raise RuntimeError('InfluxDB service must be runnint to allow data '
```
nir0s: um.. RuntimeError won't help much here. `ctx.abort_operation` please.

components/rabbitmq/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,56 @@
...
+changed = []
+for property_name in IMMUTABLE_PROPERTIES:
+    original_property = install_properties.get(property_name)
+    upgrade_property = upgrade_properties.get(property_name)
+
+    if original_property != upgrade_property:
+        changed.append(property_name)
+
+if changed:
+    raise RuntimeError(
```
nir0s: same.. and everywhere else

components/rabbitmq/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,56 @@
...
+        'RabbitMQ properties must not change during a manager '
+        'upgrade! Changed properties: {}'.format(', '.join(changed)))
```
nir0s: use explicit index formatting everywhere for py2.6 compat ({0})

components/manager/scripts/validate.py
```diff
@@ -3,9 +3,16 @@
...
 import subprocess
+from os.path import join, dirname
 
 from cloudify import ctx
 
+ctx.download_resource(
+    join('components', 'utils.py'),
+    join(dirname(__file__), 'utils.py'))
+
+import utils  # NOQA
```
tehasdf: can we import utils into this script?

nir0s: Yup. It runs remotely.

components/elasticsearch/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,59 @@
...
+
+import utils  # NOQA
+
+ES_SERVICE_NAME = 'elasticsearch'
+
+install_properties = utils.ctx_factory.get_install_properties(
+    ES_SERVICE_NAME)
+upgrade_properties = utils.ctx_factory._load_ctx_properties(
+    ES_SERVICE_NAME)
+
```
adamlavie: This would change with 'get()' after merging https://github.com/cloudify-cosmo/cloudify-manager-blueprints/pull/291/files#diff-23cb882db3b8314e49ad99212e1c4118R738

components/elasticsearch/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,59 @@
...
+    provider_context_url = urlparse.urljoin(url, 'cloudify_storage/'
+                                                 'provider_context/context')
+    try:
+        urllib2.urlopen(provider_context_url)
+    except urllib2.URLError as e:
+        ctx.abort_operation('ES returned an error when getting the provider '
+                            'context: {0}'.format(e))
+        raise
+
+elasticsearch_url = 'http://{0}:9200'.format(ES_ENDPOINT_IP)
```
adamlavie: Port should be loaded from the elasticsearch install properties

components/influxdb/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,18 @@
...
+
+import utils  # NOQA
+
+INFLUX_SERVICE_NAME = 'influxdb'
+
+utils.upgrade_validation_directories(INFLUX_SERVICE_NAME)
+
+if not utils.systemd.is_alive(INFLUX_SERVICE_NAME):
+    ctx.abort_operation('InfluxDB service must be running to allow data '
+                        'migration')
```
adamlavie: This should change and be done for all services (except for the webui maybe). This is because pretty soon the upgrade process would run run a system workflow to create a snapshot and this would require all components to be running.

components/influxdb/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
+
+from os.path import join, dirname
```
nir0s: linebreak

components/restservice/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
+
+from os.path import join, dirname
```
nir0s: linebreak

components/restservice/scripts/upgrade_validation.py
```diff
@@ -0,0 +1,18 @@
...
+from cloudify import ctx
```
nir0s: linebreak

components/utils.py
```diff
@@ -810,6 +822,11 @@ def _get_upgrade_properties(self, service_name):
         with open(upgrade_props_file) as f:
             return json.load(f)
 
+    # XXX reconcile these two with the definitions above
+    def get_install_properties(self, service_name):
```
tehasdf: We get the original inputs - from the `._get_props_file_path` file, and the upgrade inputs - from the `._load_ctx_properties` - is that right? Note that this pre-validation runs before the rollback etc files are created.
Should we use this method?

adamlavie: checkout ctx_factory.get

components/utils.py
```diff
@@ -1149,3 +1166,62 @@ def _set_upgrade_data(**kwargs):
...
+
+def upgrade_validation_directories(service_name):
+    try:
+        ctx_factory.get_install_properties(service_name)
+    except IOError:
+        ctx.abort_operation('Service {0} has no properties file'.format(
+            service_name))
+
+    if os.path.exists(
+            ctx_factory._get_rollback_properties_dir(service_name)):
```
adamlavie: since you're using the _get_rollback_properties_dir outside the factory, rename method to be publi

components/utils.py
```diff
@@ -1149,3 +1166,62 @@ def _set_upgrade_data(**kwargs):
...
+@retry((IOError, ValueError))
+def verify_http(url, predicate=None):
+    response = urllib.urlopen(url)
+    if predicate is not None and not predicate(response):
+        raise ValueError(response)
+    return True
+
+
+def changed_upgrade_properties(service_name):
+    install_properties = ctx_factory.get_install_properties(service_name)
```
adamlavie: again, ctx_factory.get should help

components/elasticsearch/scripts/creation_validation.py
```diff
@@ -0,0 +1,64 @@
...
+    try:
+        urllib2.urlopen(provider_context_url)
+    except urllib2.URLError as e:
+        ctx.abort_operation('ES returned an error when getting the provider '
+                            'context: {0}'.format(e))
+        raise
+
+
+if utils.is_upgrade:
+    install_properties = utils.ctx_factory.get_install_properties(
```
adamlavie: rebase from master. This method has been changed to be private. use get instead

components/utils.py
tehasdf: Yes, I know about `.get` - why i wanted to discuss this is, `.get` would give me this file when called with `upgrade_props=True` (because for False, it returns the rollback file, which at this point isn't even created yet).

`.get` would do what i want, but it's somewhat misleading - here i want the original install properties, so passing `upgrade_props=True` seems wrong.

btw, i suspected this might be a bug in `.get`, which is why i prototyped a method here that shows what i need, and left the `# XXX` todo comment :)

So, i should just use `.get` and don't worry about the `upgrade_props` wording?

components/manager/scripts/create.py
```diff
@@ -98,4 +98,4 @@ def normalize_agent_name(filename):
                 os.path.join(agent_archives_path, agent_id + agent_extension))
 
 
-deploy_manager_sources()
+# deploy_manager_sources()
```
nir0s: ?

components/utils.py
nir0s: Due to our time constraints, we can leave it as is and open a JIRA for it. Please change XXX to TODO

components/restservice/scripts/creation_validation.py
```diff
@@ -0,0 +1,21 @@
...
+
+ctx.download_resource(
+    join('components', 'utils.py'),
+    join(dirname(__file__), 'utils.py'))
+import utils  # NOQA
+
+
+REST_SERVICE_NAME = 'restservice'
+IMMUTABLE_PROPERTIES = [
+    'insecure_endpoints_disabled'
```
tehasdf: This really needs #175 to be merged first: then we'll check things like `verify_certificate` and agents credentials too: https://github.com/cloudify-cosmo/cloudify-manager-blueprints/pull/175/files#diff-efe92cfd0125c444af1e24a02f0454faR212

adamlavie: Upgrade doesn't support 'ssl_enabled'. use it instead of 'insecure_endpoints_disabled'

components/utils.py
adamlavie: A fix should be added in the next few days

components/utils.py
```diff
@@ -1223,3 +1223,79 @@ def verify_service_http(service_name, url, predicate=None):
         return check_http_response(url, predicate)
     except (IOError, ValueError) as e:
         ctx.abort_operation('{0} error: {1}: {2}'.format(service_name, url, e))
+
```
nir0s: are these functions used anywhere else in the blueprint? If not, why not move them to a different file and import it for each creaction_validation script?

Also, the name creaction_validation is misleading.. as it doesn't validate the creation of anything

components/riemann/scripts/creation_validation.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python
```
nir0s: The name `creation_validation` is misleading.. as you're not actually validating the creation of anything. I'd call it upgrade-validation (note the dash instead of the underscore as it's less acceptable in modules.)

WDYT?

tehasdf: They don't validate creation, but they run as part of the `cloudify.interfaces.validation.creation` interface operation. So we need to make sure it is clear that these scripts will also be run when creating, not just on upgrade (ie. so that we don't forget the `if is_upgrade:` part).

I'm all for changing the names, if you think that is fine?

components/elasticsearch/scripts/creation_validation.py
```diff
@@ -0,0 +1,63 @@
...
+def verify_properties():
+    """Compare node properties and decide if upgrading is allowed."""
+    changed = utils.changed_upgrade_properties(ES_SERVICE_NAME)
+
+    if 'es_heap_size' in changed:
+        bootstrap_heap_size, upgrade_heap_size = changed['es_heap_size']
+        bootstrap_heap_size = utils.parse_jvm_heap_size(bootstrap_heap_size)
+        upgrade_heap_size = utils.parse_jvm_heap_size(upgrade_heap_size)
+
+        if upgrade_heap_size < bootstrap_heap_size:
```
adamlavie: Please consider heap size may also end with 'b' and 'm' and 't'

components/utils.py
```diff
@@ -1223,3 +1223,79 @@ def verify_service_http(service_name, url, predicate=None):
...
+    upgrade_properties = ctx.node.properties.get_all()
+    changed = {}
+    for property_name, original_value in install_properties.items():
+        changed_value = upgrade_properties.get(property_name)
+        if original_value != changed_value:
+            changed[property_name] = (original_value, changed_value)
+    return changed
+
+
+def verify_immutable_properties(service_name, properties):
```
adamlavie: This function should consider the 'use_existing_on_upgrade' flag stated in each service's inputs

components/manager/scripts/creation_validation.py
```diff
@@ -0,0 +1,18 @@
...
+from cloudify import ctx
+
+ctx.download_resource(
+    join('components', 'utils.py'),
+    join(dirname(__file__), 'utils.py'))
+import utils  # NOQA
+
+IMMUTABLE_PROPERTIES = [
+    'ssl',
+    'security'
```
adamlavie: Keep validation on security for now. This will be removed once tested

components/utils.py
tehasdf: Should it simply use `ctx_factory._load_ctx_properties`?

nir0s: @adamlavie ?

components/riemann/scripts/creation_validation.py
nir0s: nm. Not that important

components/utils.py
tehasdf: It now short-circuits and returns empty, if `use_existing_on_upgrade` is true. Github doesn't show it in the shortened diff here (but it does in the full diff)

PR #298, opened 2016-05-16
Cfy 5301 use provided components

nir0s:


simple-manager-blueprint.yaml
```diff
@@ -92,96 +92,96 @@ inputs:
...
+      redhat_6x_agent: redhat-Santiago-agent_3.4.0-m5-b394.tar.gz
+      cloudify_windows_agent: cloudify-windows-agent_3.4.0-m5-b394.exe
 
   #############################
   # Cloudify Modules
   #############################
   rest_service_rpm_source_url:
     type: string
-    default: http://repository.cloudifysource.org/org/cloudify3/3.4.0/m5-RELEASE/cloudify-rest-service-3.4.0-m5_b394.x86_64.rpm
+    default: cloudify-rest-service-3.4.0-m5_b394.x86_64.rpm
```
nir0s: for the reviewer. the names of the changed inputs have "url" in them even though the default values are just names of files. The reason for that is that if the user provides something different other than the current defaults, it must be a url. We could solve that by providing an additional set of inputs for file names and if a "url" input is provided, ignore the name input.. but that would both take time which we currently don't have an would load up the manager on additional unwanted inputs.

components/utils.py
```diff
@@ -367,7 +372,7 @@ def __init__(self, source_path):
 
     def remove_existing_rpm_package(self):
         """
-        removes any version that satisfies the package name of the given
```
adamlavie: fix according to standard

PR #305, opened 2016-05-19
CFY-4955 rollback node properties

adamlavie:


components/rabbitmq/scripts/start.py
```diff
@@ -31,7 +31,7 @@ def set_rabbitmq_policy(name, expression, policy):
 
 if not rabbitmq_endpoint_ip:
     ctx.logger.info("Starting RabbitMQ Service...")
-    utils.systemd.start(RABBITMQ_SERVICE_NAME)
```
adamlavie: explain why we use ignore failure

components/utils.py
```diff
@@ -158,6 +158,10 @@ def move(source, destination):
     remove(source)
```
adamlavie: check if this is still required

components/utils.py
```diff
@@ -275,6 +279,12 @@ def deploy_blueprint_resource(source, destination, service_name,
                                                   user_resource=user_resource,
                                                   render=render,
                                                   load_ctx=load_ctx)
```
adamlavie: extract to function and name it well

components/utils.py
```diff
@@ -275,6 +279,12 @@ def deploy_blueprint_resource(source, destination, service_name,
...
+    if is_rollback:
+        if not resource_file:
```
adamlavie: is_file

components/utils.py
```diff
@@ -655,38 +665,28 @@ def write_to_json_file(content, file_path):
...
-def _is_rollback():
-    """Returns true if manager is in rollback state.
-    """
-    return False
+        return None
 
 
 is_upgrade = _is_upgrade()
-is_rollback = _is_rollback()
-is_install = not is_rollback and not is_upgrade
```
adamlavie: add comment to clarify

components/utils.py
```diff
@@ -655,38 +665,28 @@ def write_to_json_file(content, file_path):
...
 
 
 def _is_upgrade():
-    """Returns true if manager is in upgrade mode. This function will assume
-    manager is in upgrade state according to the maintenance mode file.
-    """
-    status_file_path = '/opt/manager/maintenance/status.json'
+    # This file is uploaded as part of the upgrade/rollback command.
+    status_file_path = '/opt/cloudify/_workflow_state.json'
+    # workflow state file will not exist during validations
```
adamlavie: remove this comment.

components/utils.py
```diff
@@ -800,15 +805,17 @@ def _get_rollback_properties_dir(self, service_name):
...
             return json.load(f)
 
-    def _get_upgrade_properties(self, service_name):
-        upgrade_props_file = self._get_props_file_path(service_name)
-        with open(upgrade_props_file) as f:
-            return json.load(f)
+    # This function should only be used when during upgrade workflow execution
+    def load_rollback_props(self, service_name):
+        upgrade_props_file = self._get_rollback_props_file_path(service_name)
+        if os.path.isfile(upgrade_props_file):
```
adamlavie: add debug message on else

components/utils.py
```diff
@@ -942,40 +960,59 @@ def _load_node_props(service_name):
...
-        if is_upgrade and os.path.isdir(service_archive_path):
-            ctx.logger.info('Archiving node {0} resources'
+        resources_dir = self._get_resources_dir(service_name)
+        if os.path.isdir(resources_dir):
+            remove(resources_dir)
+        rename(rollback_dir, resources_dir)
+
+    def _restore_service_configuration(self, rollback_dir, service_name):
+        resources_mapping = self._get_rollback_resources_json(service_name)
+        for rollback_resource, destination in resources_mapping.iteritems():
```
adamlavie: items instead of iteritems

components/utils.py
```diff
@@ -942,40 +960,59 @@ def _load_node_props(service_name):
...
+            if destination != rollback_resource:
```
adamlavie: add comment on this one

components/utils.py
```diff
@@ -1059,19 +1096,25 @@ def _create_maintenance_headers(upgrade_props=True):
...
+        config = ctx_factory.load_rollback_props('manager-config')
     security = config['security']
     security_enabled = security['enabled']
     if security_enabled:
         username = security.get('admin_username')
         password = security.get('admin_password')
         headers.update({'Authorization':
                         'Basic ' + base64.b64encode('{0}:{1}'.format(
-                            username, password))})
+                                username, password))})
```
adamlavie: why does this happen

components/utils.py
```diff
@@ -1147,5 +1190,7 @@ def _set_upgrade_data(**kwargs):
 
 # upgrade data contains info related to the upgrade process e.g  'snapshot_id'
 def _get_upgrade_data():
-    with open(UPGRADE_METADATA_FILE) as f:
-        return json.load(f)
+    if os.path.exists(UPGRADE_METADATA_FILE):
```
adamlavie: is_file

components/utils.py
adamlavie: This function actually allows me to copy the dir without having to create it first. Does this make sense?

components/utils.py
adamlavie: Added a comment instead. Extracting to method does not work out so well in this case.

PR #309, opened 2016-05-20
CFY-5216 set nginx proxy timeouts to 300s

nir0s:


components/nginx/config/nginx.conf
```diff
@@ -25,5 +25,9 @@ http {
 
     keepalive_timeout  65;
 
+    proxy_connect_timeout 300;
+    proxy_send_timeout 300;
+    proxy_read_timeout 300;
+
```
dankilman: you'll notice that I already set a 330s read timeout, in the rest config file

PR #313, opened 2016-05-24
RabbitMQ start script: add retry to verifying if the port is open.

tehasdf:
Turns out sometimes `rabbitmqctl status` can return active a bit before
rabbitmq actually starts listening. Let's retry the port accessibility
checking.


components/utils.py
```diff
@@ -1226,9 +1226,3 @@ def verify_service_http(service_name, url, predicate=None):
         return check_http_response(url, predicate)
     except (IOError, ValueError) as e:
         ctx.abort_operation('{0} error: {1}: {2}'.format(service_name, url, e))
-
-
-def verify_port_open(service_name, port, host='localhost'):
```
tehasdf: this was only used in the rabbitmq script, might as well get rid of this now that we  don't use it.

PR #316, opened 2016-05-27
Fixes to the deployment checks for security_enabled: true builds

funkyHat:


components/utils.py
```diff
@@ -1222,4 +1223,6 @@ def verify_service_http(service_name, url, predicate=None):
     try:
         return check_http_response(url, predicate)
     except (IOError, ValueError) as e:
+        if e.args[1] == 401:
```
tehasdf: is this needed? Please note that the predicate function (eg. the function that checks `.response_code in {401}`), is passed into `check_http_response` as `predicate`. It doesn't expect the response be returned from `verify_service_http` (i think currently nothing uses the return value from `verify_service_http`).
Btw, if we're to use requests, we might want to switch the IOError for `requests.exceptions.ConnectionError`?

components/utils.py
```diff
@@ -12,6 +12,7 @@
 import urllib
 import urllib2
 import hashlib
+import requests
```
nir0s: Hey. So.. you can't use libraries which are not part of the standard library as you're running scripts on a clean machine. If this worked for you, it's because of a bug we discovered like two days ago.

funkyHat:
Looks like Łukasz already fixed the `check_http_response()` issue. `check_worker_running()` for mgmtworker still needs updating to work with security enabled though.

PR #318, opened 2016-05-29
Deploying files that are rendered: overwrite if they exist (Don't merge yet)

tehasdf:
Don't merge yet, needs discussion. Will be updated or closed as needed.

When deploying a file that needs rendering, if the file already exists,
still overwrite it: the template inputs might have changed.

Without this, config files don't get overwritten on upgrade.


tehasdf:
This doesn't actually happen; i was probably running multiple upgrades, without removing the properties files between them.

PR #327, opened 2016-06-01
CFY-5381-validate-python-version

nir0s:


components/manager/scripts/validate.py
```diff
@@ -44,6 +45,23 @@ def _get_os_distro():
...
+
+
+def _validate_python_version(expected_major_version, expected_minor_version):
+    major_version, minor_version = _get_python_version()
+    ctx.logger.info('Validating Python version...')
+    if not major_version == expected_major_version or \
+            not minor_version == expected_minor_version:
+        return _error(
+            'You are currently running Python {0}.{1}. '
+            'You must be running Python 2.7 to run Cloudify Manager.'.format(
```
dankilman: if you are already passing the expected major and minor, why not format the string using them instead of hardcoding 2.7

PR #344, opened 2016-06-16
CFY-5463 fix resource download for rollback

adamlavie:


components/utils.py
```diff
@@ -928,6 +928,16 @@ def create(self, source, destination, service_name, user_resource=False,
                 self._set_resources_json(resources_props, service_name)
         return local_resource_path, destination
 
+    @staticmethod
```
nir0s:
```python
return not os.path.isfile(local_res_path) or (is_render and is_upgrade) or False
```

?

components/utils.py
```diff
@@ -928,6 +928,16 @@ def create(self, source, destination, service_name, user_resource=False,
...
+    def _is_download_required(local_res_path, is_render):
```
nir0s: why ambiguous variable name? why not `local_resource_path`?

nir0s:
Can you explain the reasoning behind this please?

components/utils.py
adamlavie: Simply for readability

PR #346, opened 2016-06-16
CFY-5469 Use `rabbitmqctl stop` in systemd ExecStop

tehasdf:
The systemd unit for rabbitmq uses commands that start and stop rabbit
in a blocking fashion - it is started/stopped before the script
returns. This is required for correct interop with systemd.

### DIFF
--- a/components/rabbitmq/config/cloudify-rabbitmq.service
+++ b/components/rabbitmq/config/cloudify-rabbitmq.service
@@ -4,11 +4,19 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-TimeoutStartSec=0
 Restart=on-failure
 EnvironmentFile=-/etc/sysconfig/cloudify-rabbitmq
+
+# avoid a race condition when creating the erlang cookie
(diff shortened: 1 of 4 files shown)


components/rabbitmq/config/cloudify-rabbitmq.service
```diff
@@ -4,11 +4,16 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-TimeoutStartSec=0
 Restart=on-failure
 EnvironmentFile=-/etc/sysconfig/cloudify-rabbitmq
+
+ExecStartPre=-/bin/sh -c "/sbin/rabbitmqctl status > /dev/null 2>&1"
```
tehasdf: This isn't strictly required for this change, but fixes a bug described in https://bugzilla.redhat.com/show_bug.cgi?id=1059913

tehasdf:
Our previous `kill-rabbit` script does the equivalent of `killall rabbitmq`, which also kills the systemd instance that is running it, resulting in a non-zero returncode from `systemctl stop cloudify-rabbitmq`

PR #351, opened 2016-06-22
CFY-4953 Preupgrade validation fixups

tehasdf:
- 'ssl' is not a separate property, it's nested under 'security'
  - rabbitmq cert properties are rabbitmq_cert_*, not broker_cert_path

### DIFF
--- a/components/manager/scripts/creation_validation.py
+++ b/components/manager/scripts/creation_validation.py
@@ -10,7 +10,6 @@ ctx.download_resource(
 import utils  # NOQA
 
 IMMUTABLE_PROPERTIES = [
-    'ssl',
     'security',
     'ssh_user'
 ]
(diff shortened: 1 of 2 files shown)


components/manager/scripts/creation_validation.py
```diff
@@ -10,7 +10,6 @@
 import utils  # NOQA
 
 IMMUTABLE_PROPERTIES = [
-    'ssl',
```
tehasdf: 'ssl' is part of 'security' already

components/rabbitmq/scripts/creation_validation.py
```diff
@@ -17,7 +17,8 @@
     'rabbitmq_password',
     'rabbitmq_endpoint_ip',
     'rabbitmq_ssl_enabled',
-    'broker_cert_path'
```
tehasdf: 'broker_cert_path' is just a path and doesn't depend on inputs at all.
To check that the certs actually haven't changed, instead use 'rabbitmq_cert_public' and 'rabbitmq_cert_private'

PR #362, opened 2016-07-07
Allow installing the DSL parser from source in mgmtworker create script

tehasdf:
The mgmtworker create script allows installing prerequisite packages
from source, like the REST service create script, but it omitted
the DSL parser package (which is required for the rest-service).


tehasdf:
Oh, the `manager_rest` package isn't actually installed, just the rieman_controller part is, which doesn't require the dsl parser. So we don't need to install it.

PR #365, opened 2016-07-11
CFY-5598 Don't pass REST credentials in cloudify_agent

tehasdf:


tehasdf:
With the other CFY-5598 changes, agents credentials are now passed in the context from the REST service. We don't need to keep the credentials in mgmtworker config, and in cloudify_agent anymore.

tehasdf:
The credentials still need to be stored in the blueprint, because they're used by the bootstrap/upgrade process. But they can't be stored in the `cloudify_agent` property because that is saved on agent hosts.

types/manager-types.yaml
```diff
@@ -153,7 +153,10 @@ node_types:
             implementation: flask_securest.authentication_providers.token:TokenAuthenticator
             properties:
               secret_key: my_secret
-              expires_in_seconds: 600
+              expires_in_seconds: 86400
```
tehasdf: = 24h

types/manager-types.yaml
```diff
@@ -153,7 +153,10 @@ node_types:
...
+          agent_rest_credentials:
```
tehasdf: should i split it in 2 properties?

types/manager-types.yaml
```diff
@@ -153,7 +153,16 @@ node_types:
...
-              expires_in_seconds: 600
+              expires_in_seconds: 86400
+          #########################################################################
+          # Agents use token authentication to communicate with the REST Service,
+          # but if token authentication is disabled, these credentials will be
+          # used instead.
+          # These credentials are also used by the scripts during bootstrap,
+          # upgrade, and rollback.
+          #########################################################################
+          rest_username: { get_input: agent_rest_username }
```
tehasdf: this can't be in `cloudify_agent` because then it would be serialized and stored on agent machines

PR #369, opened 2016-07-17
CFY-5698 Remove `--pidfile` argument from the celery call in its unit file

AviaE:
In addition, removed the `PIDFile` entry from the gunicorn unit file as
it has no effect since the `Type` entry defaults to `simple`.


AviaE:
reviewed by @ran-z

PR #380, opened 2016-07-29
CFY-5756 Added inputs to manager to allow configure default locations…

xwombat:
… of certs on agents


components/manager/scripts/configure_manager.py
```diff
@@ -66,6 +66,12 @@ def _configure_security_properties():
             agent_verify_rest_certificate
         ctx.logger.info('agent_verify_rest_certificate: {0}'.
                         format(agent_verify_rest_certificate))
+
+        # default certs paths on agents
```
dusking: remove redundant comment, code is self explanatory

inputs/manager-inputs.yaml
```diff
@@ -29,6 +29,14 @@ inputs:
     type: boolean
     default: false
 
+  local_rest_cert_file:
```
dusking: change local_rest_cert_file to **agent**_rest_cert_**path** - indicate it's the agent cert, and consistent suffix

inputs/manager-inputs.yaml
```diff
@@ -29,6 +29,14 @@ inputs:
...
+    description: Deafult location for local certificate on agents
```
dusking: update description, no need for "local": .. location of the certificate on the agent

openstack-manager-blueprint-inputs.yaml
```diff
@@ -89,6 +89,10 @@ external_network_name: ''
 
 #insecure_endpoints_disabled: false
 
+# Default locations for certificates on agents
+#local_rest_cert_file: ''
```
dusking: in the blueprint comments we have our default values, not empty string

dusking: need to update also other blueprint inputs, not just the open stack..

PR #381, opened 2016-07-31
CFY-3691 Verify certificate in HTTPS calls

tehasdf:


components/utils.py
```diff
@@ -1211,33 +1229,263 @@ def start_service(service_name, append_prefix=True, ignore_restart_fail=False):
...
+        return urlopen(request)
     except urllib2.URLError as e:
         if not should_fail:
             ctx.logger.error('Failed to {0} {1} (reason: {2})'.format(
                 method, url, e.reason))
 
 
+def rest_service_url():
+    manager_config = ctx_factory.get('manager-config')
+    security_config = manager_config['security']
```
ran-z: whats the purpose of using get in the previous line if this one assumes the value was indeed there? :)

tehasdf: note `ctx_factory` isn't a dict, it's `CtxPropertyFactory`

PR #405, opened 2016-08-25
CFY-5846-docl-integration

dankilman:


components/rabbitmq/config/rabbitmq-definitions.json
```diff
@@ -0,0 +1,22 @@
+{
```
dankilman: document external

PR #422, opened 2016-09-28
CFY-6000 Install components required by the manager HA cluster

tehasdf:
Install all the required packages. Don't configure them yet, configuration will be
on demand, enabled on an already running manager.
This just makes sure all the packages are there.


components/postgresql/scripts/create.py
```diff
@@ -41,6 +41,9 @@ def _install_postgresql():
     ctx.logger.info('Installing python libs for PostgreSQL...')
     utils.yum_install(source=psycopg2_rpm_url, service_name=PS_SERVICE_NAME)
 
+    utils.sudo('yum localinstall -y https://download.postgresql.org/pub/repos/yum/9.5/redhat/rhel-7-x86_64/pgdg-centos95-9.5-3.noarch.rpm')  # NOQA
```
tehasdf: this will need to go into the resources targz package, similar to all the other postgresql packages

PR #432, opened 2016-11-12
CFY-6090. Write log data to postgresql

jcollado:
In this PR, the logstash configuration is updated to write logs/events to PostgreSQL.

Some comments:
- For now only the message text is stored in PostgreSQL.
  - Logs are written to the `logs` table and events are written to the `events` table.
  - A more complete schema with indexing that implements full text search will be submitted in a separate PR.
- The code assumes that the following resource files are available:
  - Logstash 3rd party output plugin: [`logstash-output-jdbc-0.2.10.gem`](https://rubygems.org/downloads/logstash-output-jdbc-0.2.10.gem)
  - PostgresSQL JDBC driver: [`postgresql-9.4.1212.jar`](https://jdbc.postgresql.org/download/postgresql-9.4.1212.jar)

  My understanding is that these files need to be uploaded to an s3 bucket, but I'm not familiar with the procedure. Any advice on this would be welcome.
- The sanity check has been updated to verity some log/event has been written to PostgreSQL. However the database name has been hardcoded because adding a property to the node didn't work, that is, the property is not available to the script at runtime. Could you let me know what I'm missing in terms of how property values are passed to scripts?


components/logstash/scripts/configure.py
```diff
@@ -0,0 +1,69 @@
...
+
+CONFIG_PATH = 'components/logstash/config'
+LOGSTASH_SERVICE_NAME = 'logstash'
+
+
+def configure_logstash():
+
+    logstash_conf_path = '/etc/logstash/conf.d'
+
+    rabbitmq_username = ctx.instance.runtime_properties['rabbitmq_username']
```
idanmo: for safety, perhaps use **ctx.instance.runtime_properties.get('rabbitmq_username')**

components/logstash/scripts/create.py
```diff
@@ -71,35 +75,6 @@ def install_logstash():
...
-        '/etc/sysconfig/cloudify-logstash',
-        LOGSTASH_SERVICE_NAME)
-
-    utils.logrotate(LOGSTASH_SERVICE_NAME)
-    utils.sudo(['/sbin/chkconfig', 'logstash', 'on'])
-    utils.clean_var_log_dir(LOGSTASH_SERVICE_NAME)
-
 
-install_logstash()
+if __name__ == '__main__':
```
idanmo: Why is this needed?
I think that by default these scripts are evaluated by script plugin..

components/logstash/scripts/preconfigure_elasticsearch_properties.py
```diff
@@ -0,0 +1,17 @@
...
+    """Get elasticsearch configuration needed to generate logstash.conf."""
+    target = ctx.target
+    source = ctx.source.instance.runtime_properties
+    source['es_endpoint_ip'] = (
+        target.instance.runtime_properties['es_endpoint_ip']
+    )
+    source['es_endpoint_port'] = target.node.properties['es_endpoint_port']
+
+
+if __name__ == '__main__':
```
idanmo: Why is this needed?
I think that by default these scripts are evaluated by script plugin..

components/logstash/scripts/preconfigure_postgresql_properties.py
```diff
@@ -0,0 +1,21 @@
...
+        'postgresql_host',
+        'postgresql_db_name',
+        'postgresql_username',
+        'postgresql_password',
+    ]
+    for prop in props:
+        source[prop] = target[prop]
+
+
+if __name__ == '__main__':
```
idanmo: Why is this needed?
I think that by default these scripts are evaluated by script plugin..

components/logstash/scripts/preconfigure_rabbitmq_properties.py
```diff
@@ -0,0 +1,20 @@
...
+    props = [
+        'rabbitmq_endpoint_ip',
+        'rabbitmq_username',
+        'rabbitmq_password',
+    ]
+    for prop in props:
+        source[prop] = target[prop]
+
+
+if __name__ == '__main__':
```
idanmo: Why is this needed?
I think that by default these scripts are evaluated by script plugin..

components/manager/scripts/configure_manager.py
```diff
@@ -92,4 +92,5 @@ def main():
         _disable_requiretty()
     _configure_security_properties()
 
+
```
idanmo: ?

components/manager/scripts/sanity/get_rest_protocol.py
```diff
@@ -10,4 +10,5 @@ def get_rest_config():
     ctx.source.instance.runtime_properties['rest_protocol'] = rest_protocol
     ctx.source.instance.runtime_properties['rest_port'] = rest_port
 
+
```
idanmo: ?

components/manager/scripts/sanity/sanity.py
```diff
@@ -266,6 +278,7 @@ def perform_sanity():
                     'cleaning up sanity resources.')
     _cleanup_sanity()
 
+
```
idanmo: ?

components/elasticsearch/scripts/start.py
```diff
@@ -70,4 +70,5 @@ def main():
         # created in the 'create.py' script.
         es_upgrade_utils.restore_upgrade_data(es_endpoint_ip, ES_ENDPOINT_PORT)
 
+
```
jcollado: Not sure what version of flake8 is running in the travis CI environment, but as you can see above, the static analysis failed for 4c529cc.

components/keepalived/scripts/create.py
```diff
@@ -19,4 +19,5 @@ def install_keepalived():
     utils.yum_install(keepalived_rpm_url,
                       service_name=KEEPALIVED_SERVICE_NAME)
 
+
```
jcollado: Same answer as above, I was just trying to make flake8 in travis CI happy.

components/logstash/scripts/create.py
jcollado: In my opinion is both a matter of style and also keep the ability to test in the future. If the `if __name__ == '__main__'` pattern is not used, then it's not possible to import the script without executing the code.

components/logstash/scripts/preconfigure_elasticsearch_properties.py
jcollado: As explained above, this approach is functionally equivalent and allows to test the code in the script in the future if needed.

components/logstash/scripts/preconfigure_postgresql_properties.py
jcollado: Same as above.

components/logstash/scripts/preconfigure_rabbitmq_properties.py
jcollado: Same as above.

components/manager/scripts/sanity/get_rest_protocol.py
jcollado: Also flake8 in travis CI

components/logstash/scripts/configure.py
jcollado: Good catch. Since there's a conditional below that checks that both `rabbitmq_username` and `rabbitmq_password` are properly set, it makes sense to use `.get` to get a nice error description instead of a key error.

I'll add a new comment once I've verified and pushed the change.

jcollado: Fixed in 0d1151d

PR #433, opened 2016-11-16
CFY-6090. Fix download error

jcollado:
In this PR, the target directory for the jdbc driver file is created before curl is executed to workaround the permissions error.

### DIFF
--- a/components/logstash/scripts/create.py
+++ b/components/logstash/scripts/create.py
@@ -44,14 +44,14 @@ def install_logstash():
     ])
 
     ctx.logger.info('Installing PostgreSQL JDBC driver...')
+    jar_path = '/opt/logstash/vendor/jar'
+    jdbc_path = join(jar_path, 'jdbc')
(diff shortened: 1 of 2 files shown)


components/logstash/scripts/create.py
```diff
@@ -44,14 +44,14 @@ def install_logstash():
     ])
 
     ctx.logger.info('Installing PostgreSQL JDBC driver...')
+    jar_path = '/opt/logstash/vendor/jar'
+    jdbc_path = join(jar_path, 'jdbc')
+    utils.mkdir(jdbc_path)
+    utils.chown('logstash', 'logstash', '/opt/logstash/vendor/jar')
```
dusking: why not utils.chown('logstash', 'logstash', jar_path) ? :)

jcollado: Thanks for pointing that out. Fixed in b0f4346.

PR #435, opened 2016-11-17
CFY-6133. Store json data for logs/events in PostgreSQL

jcollado:
In this PR, the schema for the `logs` and `events` tables has been updated to include the same fields as seen in the elasticsearch records except for those fields that don't really add any information:
- `tags` is used by logstash to decide in which table to insert each record
- `type` is always set to `cloudify_log` for logs and `cloudify_event` for events

There are still a few fields that don't seem to be used that might be removed in the future:
- `message_code` seems to be always set to `null`
- `message.argument` seems to be always set to `null` (only present for events)

Even, a change in the schema might be useful to display information in a clearer way. For example, using a `message` field directly would make it easier to use instead of having a json object that contains a  `text` field in it and no other field (aside from the `argument` field mentioned above).

### DIFF
--- a/components/logstash/scripts/create.py
+++ b/components/logstash/scripts/create.py
@@ -66,10 +66,9 @@ def create_postgresql_tables():
         (
             'CREATE TABLE {0} ('
             'timestamp TIMESTAMP,'
-            'context JSONB,'
             'logger TEXT,'
             'level TEXT,'
-            'message JSONB,'
+            'message TEXT,'
             'message_code TEXT'
             ');'
             'ALTER TABLE {0} OWNER TO cloudify;'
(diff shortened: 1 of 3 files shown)


components/logstash/scripts/create.py
```diff
@@ -22,53 +22,57 @@
...
         'sudo', '-u', 'logstash',
-        '/opt/logstash/bin/plugin', 'install', logstash_output_jdbc_path,
+        '/opt/logstash/bin/plugin', 'install', plugin_path,
     ])
 
+
+def install_postgresql_jdbc_driver():
+    """Install driver used by the jdbc plugin to write data to postgresql."""
+    driver_url = (
+        'https://jdbc.postgresql.org/download/postgresql-9.4.1212.jar'
```
idanmo: Don't we have this URL set as a property in the blueprint?

jcollado: Yes, you're right. And the same thing happens with the logstash output plugin. Let me fix that.

jcollado: Fixed in 6bef3da

PR #441, opened 2016-12-02
CFY-6091. Create log tables in rest service

jcollado:
In this PR, the code that creates the `logs` and `events` tables is removed since those tables will be created based on a model defined in the rest service code.

Related: cloudify-cosmo/cloudify-manager#638


jcollado:
Looking at the test case failures it seems that while the code is expected to run in postgresql, the tests are executed using sqlite which doesn't support the tsvector type used in postgresql for the full-text search.

PR #443, opened 2016-12-12
CFY-6161 set an upper bound limit (12) for gunicorn number of workers

idanmo:
The default number of gunicorn workers (cpus*2 + 1) should have an upper bound limit as in VMs with many CPUs this can turn to a memory issue and probably not necessary to have that many gunicorn workers.
In this PR, the number of gunicorn workers is limited to 12.
The ability to configure the number of workers is preserved.

### DIFF
--- a/aws-ec2-manager-blueprint-inputs.yaml
+++ b/aws-ec2-manager-blueprint-inputs.yaml
@@ -312,7 +312,7 @@ aws_secret_access_key: ''
 #rest_service_gunicorn_worker_count:
 #  description: |
 #    The number of worker processes for handling requests.
-#    If the default value (0) is set, then (2 * cpu_count + 1 processes) will be used.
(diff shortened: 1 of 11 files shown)


jcollado:
The code looks good as it is and I tried the expression in my laptop (8 cores) and returned 12 as expected. However, I'm thinking that maybe the expression is gettig complex enough to get advantage of the environment file (`cloudify-restservice`).

What about adding something like this to the configuration file:
```
WORKER_COUNT=$(($(nproc) * 2 + 1))
MAX_WORKER_COUNT=12
```

and then update the expression in the service file (`cloudify-restservice.service`) with something like this:
```
 -w {{ node.properties.gunicorn_worker_count or '$((WORKER_COUNT > MAX_WORKER_COUNT ? MAX_WORKER_COUNT : WORKER_COUNT))' }} \
```

idanmo:
Good idea!

PR #455, opened 2016-12-22
CFY-6091. Insert deployment reference in both logs and events

jcollado:
In this PR, a reference to the deployment is added to both the events and the logs tables since that information needs to be included in the response from the API.

Related: cloudify-cosmo/cloudify-manager#656

### DIFF
--- a/components/logstash/config/logstash.conf
+++ b/components/logstash/config/logstash.conf
@@ -55,8 +55,9 @@ output {
             driver_class => 'org.postgresql.Driver'
             connection_string => 'jdbc:postgresql://{{ ctx.instance.runtime_properties.postgresql_host }}:5432/{{ ctx.instance.runtime_properties.postgresql_db_name }}?user={{ ctx.instance.runtime_properties.postgresql_username }}&password={{ ctx.instance.runtime_properties.postgresql_password }}'
             statement => [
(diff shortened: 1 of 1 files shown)


components/logstash/config/logstash.conf
```diff
@@ -55,8 +55,9 @@ output {
             driver_class => 'org.postgresql.Driver'
             connection_string => 'jdbc:postgresql://{{ ctx.instance.runtime_properties.postgresql_host }}:5432/{{ ctx.instance.runtime_properties.postgresql_db_name }}?user={{ ctx.instance.runtime_properties.postgresql_username }}&password={{ ctx.instance.runtime_properties.postgresql_password }}'
             statement => [
-              "INSERT INTO logs (timestamp, execution_fk, logger, level, message, message_code) VALUES(CAST (? AS TIMESTAMP), (SELECT storage_id FROM executions WHERE id = ?), ?, ?, ?, ?)",
+              "INSERT INTO logs (timestamp, deployment_fk, execution_fk, logger, level, message, message_code) VALUES(CAST (? AS TIMESTAMP), (SELECT storage_id FROM deployments WHERE id = ?), (SELECT storage_id FROM executions WHERE id = ?), ?, ?, ?, ?)",
```
idanmo: In the corresponding PR in cloudify-manager I asked why deployment_fk is needed is you this information is available through executions_fk..?

jcollado: Yes, there's no real need to have that field. I was just trying to make queries simpler, but now they've been updated to be based on the `execution_fk` instead of the `deployment_fk`.

jcollado:
Closing this PR, since there's no need to have the `execution_fk` field in the event tables.

PR #467, opened 2017-01-16
CFY-6307-Manager-blueprints-inputs-should-not-include-default-package…

limor-gs:
…s-and-urls

### DIFF
--- a/components/utils.py
+++ b/components/utils.py
@@ -1133,8 +1133,24 @@ class BlueprintResourceFactory(object):
     def _download_source_resource(source, local_resource_path):
         is_url = source.startswith(('http', 'https', 'ftp'))
         filename = get_file_name_from_url(source) if is_url else source
-        local_filepath = os.path.join(CLOUDIFY_SOURCES_PATH, filename)
+
         is_manager_package = filename.startswith('cloudify-manager-resources')
+        local_filepath = os.path.join(CLOUDIFY_SOURCES_PATH, filename)
+        # if is_manager_package:
+        #     local_filepath = os.path.join(CLOUDIFY_SOURCES_PATH, filename)
+        # else:
+        #     ctx.logger.info("limor - file name: {0}".format(filename))
+        #     ctx.logger.info("limor - folder name: {0}".format(CLOUDIFY_SOURCES_PATH))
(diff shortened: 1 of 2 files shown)


components/utils.py
```diff
@@ -520,6 +521,22 @@ def yum_install(source, service_name):
...
+    local_filepath_list = \
+        [fn for fn in glob.glob(os.path.join(CLOUDIFY_SOURCES_PATH, filename))
+         if not os.path.basename(fn).startswith(SINGLE_TAR_PREFIX)]
+    if not local_filepath_list:
+        ctx.abort_operation("File: {0} does not exist in sources path: {1}".
+                            format(filename, CLOUDIFY_SOURCES_PATH))
+    if len(local_filepath_list) > 1:
+        ctx.abort_operation("More than one file: {0} found in sources path:"
+                            " {1}".format(filename, CLOUDIFY_SOURCES_PATH))
+    local_filepath = ''.join(local_filepath_list)
```
dusking: I think that local_filepath_list[0] is more explicit

components/utils.py
```diff
@@ -1133,8 +1150,14 @@ def _download_resource_and_render(self, source, dest, service_name,
     def _download_source_resource(source, local_resource_path):
         is_url = source.startswith(('http', 'https', 'ftp'))
         filename = get_file_name_from_url(source) if is_url else source
-        local_filepath = os.path.join(CLOUDIFY_SOURCES_PATH, filename)
-        is_manager_package = filename.startswith('cloudify-manager-resources')
+
```
dusking: blank line..

components/utils.py
```diff
@@ -1158,7 +1181,18 @@ def _load_node_props(service_name):
 
     def _get_local_file_path(self, service_name, resource_name):
         base_service_res_dir = self.get_resources_dir(service_name)
-        dest_file_path = os.path.join(base_service_res_dir, resource_name)
+
+        if (resource_name.startswith('cloudify')
```
dusking: please move this complicated condition into a function with explanatory name and comments..

components/utils.py
```diff
@@ -1158,7 +1181,18 @@ def _load_node_props(service_name):
...
+            or resource_name.find('-agent_') != -1) \
```
tehasdf: `'-agent_' not in resource_name` is nicer than using `.find` (with its -1 weirdness)

PR #473, opened 2017-01-25
Cfy 6349 set ip manager on image boot

idanmo:


components/manager-ip-setter/scripts/create.py
```diff
@@ -0,0 +1,42 @@
...
+             'manager-ip-setter',
+             'scripts',
+             MANAGER_IP_SETTER_SCRIPT_NAME),
+        config_file_temp_destination)
+    utils.move(config_file_temp_destination, MANAGER_IP_SETTER_SCRIPT_PATH)
+    utils.chmod('+x', MANAGER_IP_SETTER_SCRIPT_PATH)
+    utils.systemd.configure(MANAGER_IP_SETTER_SERVICE_NAME)
+
+
+if os.environ.get('set_manager_ip_on_boot').lower() == 'true':
```
tehasdf: the `.get` is ineffective because missing values return None, so it would TypeError on `.lower()`; use 2-arg `.get`

components/manager-ip-setter/scripts/manager-ip-setter.sh
```diff
@@ -0,0 +1,81 @@
...
+  echo "Updating logstash.conf.."
+  /usr/bin/sed -i -e "s/host => "'"'".*"'"'"/host => "'"'"${ip}"'"'"/" /etc/logstash/conf.d/logstash.conf
+
+  echo "Updating broker_config.json.."
+  /usr/bin/sed -i -e "s/"'"'"broker_hostname"'"'": "'"'".*"'"'"/"'"'"broker_hostname"'"'": "'"'"${ip}"'"'"/" /opt/mgmtworker/work/broker_config.json
+
+  echo "Updating broker_ip in provider context.."
+
+  python << END
+import requests, json, sys, time
```
tehasdf: this is just using the system python - can we guarantee it has requests?

components/manager-ip-setter/config/cloudify-manager-ip-setter.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Cloudify Manager IP Setter
+Before=cloudify-mgmtworker.service cloudify-amqpinflux.service cloudify-riemann.service logstash.service
+After=network.target cloudify-restservice.service
+
+[Service]
```
tehasdf: maybe should be a `Type=oneshot`? does systemd try to restart the script after it exits?

idanmo: It runs only once.
The behavior of Type=simple and ExecStart=something is similar to oneshot.

PR #477, opened 2017-01-31
CFY-6373 Security groups: allow cluster-related ports

tehasdf:


aws-ec2-manager-blueprint.yaml
```diff
@@ -275,6 +275,27 @@ node_templates:
           from_port: 53229
           to_port: 53229
           src_group_id: { get_property: [agents_security_group, resource_id] }
+        # Cluster-related ports, only used in a manager cluster
+        - ip_protocol: tcp
+          from_port: 8300
+          to_port: 8300
+          src_group_id: { get_property: [agents_security_group, resource_id] }
```
tehasdf: aws blueprint has no management security group (like openstack blueprint does), we should just use the agents security group?

PR #499, opened 2017-02-27
CFY-6497 generate internal SSL certificates on boot

idanmo:
In this PR, the manager-ip-setter service is updated to generate the internal SSL certificates when booting the manager from an image.
In order to achieve that, the utils.generate_internal_ssl_cert function is used.


components/manager-ip-setter/scripts/manager-ip-setter.sh
```diff
@@ -28,6 +28,11 @@ function set_manager_ip() {
   echo "Updating broker_ip in provider context.."
   /opt/cloudify/manager-ip-setter/update-provider-context.py ${ip}
 
+  echo "Creating internal SSL certificates.."
+  rm -f /etc/cloudify/ssl/cloudify_internal_*.pem
```
mcouthon: Remove using utils constants

PR #501, opened 2017-03-01
CFY-6492 Make the file server always secured

mcouthon:
* The file server is now always internally accessed via the REST's port (with a /resources suffix), and is then redirected to  its internal port
 * All the folder names now reside in cloudify-plugins-common.constants (seeing as how they are indeed constants)
 * All the workers (agent + mgmt) now need to have much fewer params regarding the file server


components/nginx/config/https-file-server.cloudify
```diff
@@ -2,15 +2,15 @@
...
+    server_name         127.0.0.1;
 
     # force http redirect to https
     if ($scheme = http) {
         return 301 https://$server_name$request_uri;
     }
 
-    ssl_certificate     {{ ctx.target.instance.runtime_properties.external_cert_path }};
-    ssl_certificate_key {{ ctx.target.instance.runtime_properties.external_key_path }};
+    ssl_certificate     {{ ctx.source.instance.runtime_properties.internal_cert_path }};
```
tehasdf: also do this in `https-external-rest-server` :)

mcouthon: Wait, what am I missing? You *don't* need the key in nginx? In every example I saw both cert and key were provided.

tehasdf: you do; i'm just reminding you to also do this change (`target` -> `source`) inside `https-external-rest-server.cloudify` (another file which needs this same change)

actually, i sent a PR which includes these changes (both files) and another typo fix as well #505

PR #504, opened 2017-03-02
CFY-6507 add config for encoding id for api_token

dusking:


components/restservice/scripts/configure.py
```diff
@@ -34,13 +35,25 @@
 REST_SERVICE_NAME = 'restservice'
 
 
+def _random_alphanumeric(result_len=31):
+    """
+    :return: random string of unique alphanumeric characters
+    """
+    ascii_alphanumeric = (range(48, 58) + range(65, 91) + range(97, 123))
+    random.shuffle(ascii_alphanumeric)
+    return reduce(lambda i, s: i + chr(s), ascii_alphanumeric[:result_len], "")
```
tehasdf: you can also use `ascii_alphanumeric = string.ascii_letters + string.digits` and then `return ''.join(random.sample(ascii_alphanumeric, result_len))` - no reduce(), no going through integers, py2-and-3-compatible :)

dusking: cool, thanks

PR #505, opened 2017-03-03
Fix typos in passing the external cert and key to nginx config

tehasdf:


components/nginx/config/https-external-rest-server.cloudify
```diff
@@ -11,8 +11,8 @@ server {
       return 301 https://$server_name$request_uri;
   }
 
-  ssl_certificate     {{ ctx.target.instance.runtime_properties.external_cert_path }};
-  ssl_certificate_key {{ ctx.target.instance.runtime_properties.external_key_path }};
+  ssl_certificate     {{ ctx.source.instance.runtime_properties.external_cert_path }};
```
tehasdf: fwiw the code that sets it on source is here: https://github.com/cloudify-cosmo/cloudify-manager-blueprints/blob/master/components/nginx/scripts/preconfigure.py#L111-L112

(in case we decide that `.target` is what we want after all)

tehasdf:
done in #501

PR #506, opened 2017-03-03
Nginx: use 308 for the http-to-https redirect

tehasdf:
Using 301 makes requests retry the same request, but without a body;
use 308 so that when it tries the new location, it also sends
the body again.


tehasdf:
done inside #545

PR #517, opened 2017-03-15
CFY-6562 Generate external ssl cert if not provided - bugfix

tehasdf:
The error message isn't actually in `str(e)` but it is in `e.stderr`


tehasdf:
done in #519

PR #520, opened 2017-03-19
CFY-6582 Run manager-ip-setter scripts using python from manager venv

mcouthon:


geokala:
My last PR actually made it use the mgmtworker venv, and copied the utils over, so I think this is probably not needed now?

mcouthon:
@geokala Absolutely.

PR #521, opened 2017-03-20
CFY-6565 Set stage to restart=always with a 10 sec delay

tehasdf:


components/stage/config/cloudify-stage.service
```diff
@@ -5,7 +5,8 @@ After=network-online.target
 
 [Service]
 TimeoutStartSec=0
-Restart=on-failure
+Restart=always
```
tehasdf: need to do `always` because stage exits with a 0 error code on db connection failures

PR #542, opened 2017-04-08
CFY-3532- Non-root management worker service

geokala:


geokala:
Tests require https://github.com/cloudify-cosmo/cloudify-system-tests/pull/514

PR #549, opened 2017-04-16
Cfy 6734+6736 non root components

geokala:
Mgmtworker and manager IP setter

Includes some changes to improve IP setter.

### DIFF
--- a/components/manager-ip-setter/scripts/create.py
+++ b/components/manager-ip-setter/scripts/create.py
@@ -15,6 +15,10 @@ import utils  # NOQA
 
 MANAGER_IP_SETTER_SERVICE_NAME = 'manager-ip-setter'
 ctx_properties = utils.ctx_factory.create(MANAGER_IP_SETTER_SERVICE_NAME)
+MANAGER_IP_SETTER_USER = ctx_properties['os_user']
+MANAGER_IP_SETTER_GROUP = ctx_properties['os_group']
+HOMEDIR = ctx_properties['os_homedir']
+SUDOERS_INCLUDE_DIR = ctx_properties['sudoers_include_dir']
 
 MANAGER_IP_SETTER_SCRIPT_NAME = 'manager-ip-setter.sh'
 UPDATE_PROVIDER_CONTEXT_SCRIPT_NAME = 'update-provider-context.py'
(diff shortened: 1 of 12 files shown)


components/manager-ip-setter/scripts/create.py
```diff
@@ -15,30 +15,61 @@
...
 ctx_properties = utils.ctx_factory.create(MANAGER_IP_SETTER_SERVICE_NAME)
+MANAGER_IP_SETTER_USER = ctx_properties['os_user']
+MANAGER_IP_SETTER_GROUP = ctx_properties['os_group']
+HOMEDIR = ctx_properties['os_homedir']
+SUDOERS_INCLUDE_DIR = ctx_properties['sudoers_include_dir']
 
 MANAGER_IP_SETTER_SCRIPT_NAME = 'manager-ip-setter.sh'
 UPDATE_PROVIDER_CONTEXT_SCRIPT_NAME = 'update-provider-context.py'
-CREATE_INTERNAL_SSL_CERTS_SCRIPT_NAME = 'create-internal-ssl-certs.py'
+CREATE_SSL_CERTS_SCRIPT_NAME = 'create-ssl-certs.py'
```
idanmo: Why did you rename this?

components/manager-ip-setter/scripts/manager-ip-setter.sh
```diff
@@ -19,17 +19,27 @@ function set_manager_ip() {
...
 
   echo "Creating internal SSL certificates.."
-  /opt/cfy/embedded/bin/python /opt/cloudify/manager-ip-setter/create-internal-ssl-certs.py ${ip}
+  /opt/mgmtworker/env/bin/python /opt/cloudify/manager-ip-setter/create-ssl-certs.py ${ip}
   
-  echo "Restarting nginx.."
+  echo "Restarting services.."
+  # Restarting all (except postgres) to avoid issues with not correctly reloading SSL certs
   systemctl restart nginx
+  systemctl restart cloudify-rabbitmq
```
idanmo: Why is it needed to restart all these services?
RabbitMQ already appears in line 45..

components/rabbitmq/config/rabbitmq-env.conf
```diff
@@ -0,0 +1 @@
+NODENAME=cloudify-manager@localhost
```
idanmo: Why is this needed?

components/rabbitmq/scripts/create.py
```diff
@@ -81,6 +81,18 @@ def _install_rabbitmq():
...
+    # a manager is deployed from an image but given a new hostname.
+    # This is likely to cause problems with clustering of rabbitmq if this is
+    # done at any point, so at that point a change to the file and cleaning of
+    # mnesia would likely be necessary.
+    utils.deploy_blueprint_resource(
+        '{0}/rabbitmq-env.conf'.format(CONFIG_PATH),
+        '/etc/rabbitmq/rabbitmq-env.conf',
+        RABBITMQ_SERVICE_NAME)
+    # Delete old mnesia node
+    utils.sudo(['rm', '-rf', '/var/lib/rabbitmq/mnesia'])
```
idanmo: why would this exist?

types/manager-types.yaml
```diff
@@ -807,6 +819,22 @@ node_types:
...
+        default: cfyipsetter
+      os_group:
+        description: IP setter's OS group. Will be created if not existing, and used to run the service.
+        type: string
+        default: cfyipsetter
+      os_homedir:
+        description: IP setter's OS user's home directory. Will not be created if missing.
+        type: string
+        default: /etc/cloudify
+      sudoers_include_dir:
```
idanmo: In general, please remember that if more users should be added to sudoers file, we should probably think of a better approach.. maybe as part of some setup node or some other idea.

components/manager-ip-setter/scripts/create.py
geokala: Because it is now regenerating the non-internal cert too.

components/manager-ip-setter/scripts/manager-ip-setter.sh
geokala: Whoops, I thought I removed the restarts from elsewhere.

I ran into issues on one of the build attempts where I had to restart some services that weren't restarted because at least one appeared to be caching a cert, so I put in restarts on all of the services we use so that even if we add extra logic to the IP setting at a later date, things should just work (tm), rather than having to decide specifically which services need restarting each time.

I'll remove the earlier RabbitMQ and retest.

components/rabbitmq/config/rabbitmq-env.conf
geokala: Without it, deploying the manager image with a different hostname to the one it had originally was causing rabbit to fail to start because it couldn't find the node (default is something like rabbit@<hostname>, and <hostname> was different.)

components/rabbitmq/scripts/create.py
geokala: Because rabbit has already been started at this point, and not removing it will leave the old named mnesia datastore.

types/manager-types.yaml
geokala: Potentially, yes. I'm not sure how many we're going to need to add on this run though, so adding extra complexity and retrieving the sudoers include dir path from a relationship is probably overkill until we add another one?

PR #550, opened 2017-04-17
CFY 6745. Use haveged

jcollado:
Since it seems that `rngd -r /dev/urandom` is not really a good practice, in this PR `haveged` is used instead.

Also, the `haveged` service is started/stopped before/after installation of the logstash plugins. This doesn't guarantee that entropy generated by `haveged` is not used later for some other things, but still this should be a safer approach than leaving `haveged` running.

Note that these changes expect to find the `haveged` rpm in the cloudify manager resources file, so this PR won't be merged until the new resources files has been generated.

### DIFF
--- a/components/logstash/scripts/create.py
+++ b/components/logstash/scripts/create.py
@@ -84,10 +84,9 @@ def install_logstash():
 
     utils.yum_install(logstash_source_url, service_name=LOGSTASH_SERVICE_NAME)
 
-    # Make sure there's enough entropy in /dev/random
-    # before installing plugins
-    utils.run(['sudo', 'yum', 'install', '-y', 'rng-tools'])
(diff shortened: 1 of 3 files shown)


geokala:
Worth reading: https://security.stackexchange.com/questions/34523/is-it-appropriate-to-use-haveged-as-a-source-of-entropy-on-virtual-machines

It's notable that there's not a strong consensus on how useful this is for crypto purposes. We are likely to be better changing the way we run the jruby commands to force them to use /dev/urandom rather than changing the approach used by the image to generate random numbers, especially given that this may alter the way a user has deliberately configured their system for their own security/auditing reasons.

jcollado:
Closing this PR in favor of #551 

@geokala Thanks for your feedback

PR #552, opened 2017-04-18
CFY-6737- Run riemann as non-root

geokala:
Not to be merged until docl tests done (comment will be added to note this).


components/riemann/scripts/create.py
```diff
@@ -56,6 +65,9 @@ def install_riemann():
     utils.mkdir(riemann_config_path)
     utils.mkdir('{0}/conf.d'.format(riemann_config_path))
 
+    # utils.chown cannot be used as it will change both user and password
```
idanmo: user and group?

PR #553, opened 2017-04-18
CFY-3532 CFY-6735 non root restservice

funkyHat:


components/restservice/config/cloudify-restservice.service
```diff
@@ -4,11 +4,13 @@ Wants=network-online.target dbus.service
...
 
 [Service]
+User={{ node.properties.service_user }}
+Group={{ node.properties.service_user }}
 TimeoutStartSec=0
 Restart=on-failure
 EnvironmentFile=-/etc/sysconfig/cloudify-restservice
 ExecStart=/bin/sh -c '/opt/manager/env/bin/gunicorn \
-    --pid /var/run/gunicorn.pid \
+    --pid /run/cloudify-restservice/pid \
```
idanmo: why /pid and not .pid?

inputs/manager-inputs.yaml
```diff
@@ -62,6 +62,12 @@ inputs:
     type: integer
     default: 20
 
+  rest_service_system_user:
```
idanmo: I'm not sure we want to expose these inputs at the moment.

types/manager-types.yaml
```diff
@@ -636,6 +636,9 @@ node_types:
         description: Number of log files to keep when rotating.
         type: integer
         default: { get_input: rest_service_log_files_backup_count }
+      service_user:
```
idanmo: this should probably be consistent with other user/group properties... os_user, os_group etc...

components/restservice/scripts/configure.py
```diff
@@ -57,6 +57,13 @@ def _deploy_security_configuration():
         'encoding_min_length': 5
     }
 
+    rest_service_user = ctx.node.properties['service_user']
```
idanmo: os_user?

components/restservice/config/cloudify-restservice.service
funkyHat: Because `cloudify-restservice` doesn't have permission to create files in `/run`

idanmo:
How about adding an `os_group` property for being consistent with other services?

PR #554, opened 2017-04-19
Stop IP setter errors due to startup order

geokala:


components/manager-ip-setter/scripts/manager-ip-setter.sh
```diff
@@ -35,11 +35,13 @@ function set_manager_ip() {
...
   systemctl restart nginx
-  systemctl restart cloudify-rabbitmq
   systemctl restart cloudify-amqpinflux
   systemctl restart cloudify-influxdb
-  systemctl restart cloudify-restservice
   systemctl restart cloudify-mgmtworker
+  systemctl restart cloudify-rabbitmq
+  systemctl restart cloudify-restservice
+  systemctl restart cloudify-riemann
+  systemctl restart cloudify-stage
```
idanmo: why do we care about stage here?

geokala: I just put them all in to save on any present or future issues related to any files we change (e.g. if stage was reading and caching the internal public cert or any IPs then this would just make that problem go away before it became an issue).

idanmo: The problem is that in community-edition stage does not exist..
So I guess it would be better to remove it or check if it exists before attempting to restart it..

geokala: Ah. Yeah, I'll put in an || true

PR #556, opened 2017-04-22
Fix snapshot permissions

geokala:
Also requires system workflows change- PR on the way.

### DIFF
--- a/components/mgmtworker/scripts/configure.py
+++ b/components/mgmtworker/scripts/configure.py
@@ -32,6 +32,7 @@ CONFIG_PATH = 'components/{0}/config'.format(SERVICE_NAME)
 ctx_properties = utils.ctx_factory.create(SERVICE_NAME)
 MGMTWORKER_USER = ctx_properties['os_user']
 MGMTWORKER_GROUP = ctx_properties['os_group']
+HOMEDIR = ctx_properties['os_homedir']
 
 
 def configure_mgmtworker():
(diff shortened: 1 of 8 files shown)


components/mgmtworker/scripts/configure.py
```diff
@@ -75,5 +71,18 @@ def configure_logging():
     utils.move(config_file_temp_destination, config_file_destination)
 
 
+def prepare_snapshot_permissions():
+    pgpass_location = '/root/.pgpass'
+    destination = join(HOMEDIR, '.pgpass')
+    utils.chmod('400', pgpass_location)
+    utils.chown(MGMTWORKER_USER, MGMTWORKER_GROUP, pgpass_location)
+    utils.sudo(['mv', pgpass_location, destination])
+    utils.sudo(['chgrp', MGMTWORKER_GROUP, '/opt/manager'])
```
idanmo: Write access is needed to this folder or a sub folder within /opt/manager?

geokala:
Requires: https://github.com/cloudify-cosmo/cloudify-manager/pull/815

components/mgmtworker/scripts/snapshot_permissions_fixer
```diff
@@ -0,0 +1,5 @@
+#! /usr/bin/env bash
+restservice_user=$(cat /usr/lib/systemd/system/cloudify-restservice.service | grep User | sed 's/User=//')
```
idanmo: Wouldn't it be better to render this file on bootstrap with the right values?

components/mgmtworker/scripts/configure.py
geokala: This folder specifically- it's where it creates a snapshot_credentials (IIRC) folder. I'm not sure why it creates it, but it gets grumpy if it can't.

components/mgmtworker/scripts/snapshot_permissions_fixer
geokala: Potentially- that'd require putting in extra relationships and pulling the info from there, and rather than doing that I figured this would probably be sensible since it can't get out of date even if someone slips up and makes an upgrade not work precisely as expected in future.

idanmo: But you still need to make sure the cloudify-restservice.serivce file exists before running the scripts.. so you need to have a relationship in order to make the order right (if its not already like that..).

geokala: This is true, and that relationship... apparently doesn't exist. Update coming shortly.

PR #572, opened 2017-04-30
CFY-6809 Unify restservice, mgmtworker and manager-ip-setter users

mcouthon:


funkyHat:
I'm concerned that this change is a move in the wrong direction. Please see comments in JIRA
(the change to combine all sudoers entires into a single file seems good by itself though)

PR #576, opened 2017-05-03
CFY-6800 Allow sudo systemctl for user cloudify-restservice

inbala:


geokala:
Seconded-
systemctl   edit NAME...                    Edit one or more unit files
followed by:
systemctl restart NAME

That allows any arbitrary command execution. If restservice needs to be able to restart specific things (for instance) then they should be put into a script and that script should be called with sudo by the restservice. Otherwise we've basically just spent a lot of effort to pretend we're not running as root.

inbala:
Right, it has to be done for now but we should revisit.

PR #582, opened 2017-05-04
CFY 6093. rabbithole as logstash alternative

jcollado:
This PR is a proof of concept to add rabbithole, a new node type that can be used to replace logstash to store events and logs in PostgreSQL. In fact, there is a new blueprint, `simple-manager-blueprint-rabbithole.yaml`, that can be used to bootstrap a new manager and work with it as usual.

### DIFF
--- a/components/rabbithole/config/config.yml
+++ b/components/rabbithole/config/config.yml
@@ -1,0 +1,52 @@
+size_limit: 5
+time_limit: 15
+blocks:
+  - name: input
+    type: amqp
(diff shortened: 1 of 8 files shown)


jcollado:
Since this is an experiment, another way to install rabbithole in a system in which cloudify is already running without having to bootstrap a new one would be to use the ansible playbook here:
https://github.com/jcollado/rabbithole-ansible-cloudify

PR #594, opened 2017-05-18
Disable SSLv3

isaac-s:


isaac-s:
The version of nginx we use with 4.x already disabled SSLv3 by default. Closing.

PR #596, opened 2017-05-22
CFY-6908 Add cfyuser to the stage group

tehasdf:
To allow restoring stage data from snapshots, the mgmtworker
user needs to be able to write to the stage directories


tehasdf:
obsoleted by #597 - instead of adding cfyuser to the stage group, we use a script that will be run as the stage user via sudo

PR #604, opened 2017-06-01
CFY-6960: provide a way to override /tmp for executables

isaac-s:


isaac-s:
Fixed the PG startup script so it doesn't assume `/tmp`. This required adding a function to `utils.py` that is similar to the `get_exec_tempdir` in `cloudify-plugins-common`. I dislike copying code but in the manager blueprints' case there's no other choice at the moment.

PR #613, opened 2017-06-26
Support non-IPs in cert generation

geokala:


components/utils.py
```diff
@@ -283,11 +283,38 @@ def _generate_ssl_certificate(ips,
...
-    cert_metadata = '{0},DNS:localhost'.format(
-        ','.join(metadata_items))
+    altnames = set(ips)
+
+    # Ensure we trust localhost
+    altnames.add('127.0.0.1')
+    altnames.add('localhost')
+
+    subject_altdns = [
+        'DNS:{name}'.format(name=name)
```
tehasdf: note that we also add ips as `DNS:ip.ip.ip.ip` because chrome needs a DNS entry even with an ip, just an IP: entry won't  be enough

PR #621, opened 2017-07-04
move composer web port to 8443. http and https if SSL enabled

maimon33:


components/nginx/config/https-external-rest-server.cloudify
```diff
@@ -31,4 +45,18 @@ server {
   # to add the header manually.
   add_header Location https://$host$request_uri always;
   return 308;
-}
\ No newline at end of file
+}
+
+# force http redirect to https
+server {
```
tehasdf: this block won't work because the same port is used for the actual HTTPS server above. The first block will catch the plain HTTP request and return a 400 error saying that HTTPS should be used instead.

Instead, if we want to redirect http-to-https on the same port, something like this directive can be put into the above block (into the `ssl` composer block):
```
error_page 497 =307 https://$host:$server_port$request_uri;
```
497 is nginx's "meta" error code that means "http request was sent to a https port".
With this directive, requesting http://ip:8443, does return a 307 with `Location: https://ip:8443`

PR #623, opened 2017-07-06
restart composer when starting from image

maimon33:


components/manager-ip-setter/scripts/manager-ip-setter.sh
```diff
@@ -45,6 +45,9 @@ function set_manager_ip() {
   if $(systemctl list-units | grep cloudify-stage > /dev/null); then
       systemctl restart cloudify-stage
   fi
+  if $(systemctl list-units | grep cloudify-composer> /dev/null); then
```
idanmo: Add space before ">"

PR #624, opened 2017-07-14
CFY-7073: provide way to control max gate keeper independently of agent config

isaac-s:
The current limitation makes no sense. I believe it's a remant from the days when we used to have deployment workers. It's irrelevant anymore. By default, the manager should be configured to allow for maximum parallelism, constrained only by available hardware resources.

Maybe we should consider allowing an input to disable the gatekeeper altogether.


inputs/manager-inputs.yaml
```diff
@@ -436,6 +436,14 @@ inputs:
     type: integer
     default: 2
 
+  gate_keeper_bucket_size:
```
idanmo: Why keep this uninformative name? what about `max_parallel_operations_per_deployment` ?

isaac-s:
@idanmo changed input name. Makes sense.

PR #625, opened 2017-07-14
Skip redirect-to-https for the cloudify_agent upgrade url

tehasdf:


tehasdf:
Travis is red because it doesn't seem to actually use the newest master, even though this branch is based on it.

PR #630, opened 2017-07-20
Store manager version and build time

tehasdf:


inputs/manager-inputs.yaml
```diff
@@ -1,4 +1,9 @@
 inputs:
+  #############################
+  # Version Label
+  #############################
+  manager_version:
+    default: 4.2.0-.dev1
```
mcouthon: Isn't it weird for the version to appear in the inputs?

PR #635, opened 2017-07-31
CFY-7215 Use storage_manager directly in ipsetter

tehasdf:
Instead of using the REST API to update provider context, simply use
restservice's storage_manager directly - this makes it faster, and avoids
a circular dependency issue (where restservice needs to be started for
ipsetter to work, but without ipsetter, restservice is using the wrong cert)


tehasdf:
taking bets on how much stuff will be broken by this :)
unfortunately we don't currently have a way to run the whole test suite with thisCloudify 4.2 Sprint 4

PR #643, opened 2017-08-26
Added option to provide additional IPs / DNS names to internal and ex…

isaac-s:
…ternal certs

### DIFF
--- a/components/utils.py
+++ b/components/utils.py
@@ -272,6 +272,23 @@ def remove(path, ignore_failure=False):
     sudo(['rm', '-rf', path], ignore_failures=ignore_failure)
 
 
+def _get_cert_ips_and_dns(ips, dns_names=None):
+    assert ips is not None
+    assert len(ips) > 0
+
+    # Remove duplicates from ips
+    cert_ips = set(ips)
+    cert_ips.add('127.0.0.1')
+    cert_dns = set(dns_names) if dns_names else set()
+    # Add all IP's as DNS's as well. See:
+    # http://www.michaelm.info/blog/?p=1281
+    # https://bugs.chromium.org/p/chromium/issues/detail?id=72726
+    cert_dns.update(cert_ips)
+    cert_dns.add('localhost')
+
+    return cert_ips, cert_dns
+
+
 def _generate_ssl_certificate(ips,
                               dns_names,
                               cn,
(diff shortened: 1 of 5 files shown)


components/utils.py
```diff
@@ -337,17 +340,19 @@ def _generate_ssl_certificate(ips,
     return cert_path, key_path
 
 
-def generate_internal_ssl_cert(ip):
+def generate_internal_ssl_cert(ip, additional_ips=[], dns_names=[]):
```
tehasdf: please note that using mutable default arguments for functions is bad style in python, because they are only created once, and each function call will get the same list object. You are appending to `dns_names`, so for the first call, this function would receive `dns_names=[]`, but for the second call, it would receive `dns_names=['localhost']` (and for the third, `dns_names=['localhost', 'localhost']` etc).
This is only called once so the bug isn't visible, but it's better to guard against that.
The common idiom is to do in the function header `dns_names=None` and then inside the function body `if dns_names is None: dns_names = []`

(btw, i thought our linter checked for that?)

components/utils.py
```diff
@@ -286,9 +287,11 @@ def _generate_ssl_certificate(ips,
     # Remove duplicates from ips
     ips.append('127.0.0.1')
     ips = set(ips)
-    metadata_items = ['IP:{0},DNS:{0}'.format(x) for x in ips]
-    cert_metadata = '{0},DNS:localhost'.format(
-        ','.join(metadata_items))
+    dns_names.append('localhost')
+    dns_names = set(dns_names)
+    metadata_ips = ['IP:{0}'.format(x) for x in ips]
```
tehasdf: the reason we used to put the ip addreses as `DNS:` entries is that, some browsers (chrome) basically outright ignore `IP` entries, but work fine when the ip addresses are in `DNS:` entries (surprisingly). (eg. http://www.michaelm.info/blog/?p=1281 - the bug linked there is already closed but AFAIR it did still work like this when I checked this year. I'll check again)

So if this is used to generate a cert that can be seen by browsers (and it is used for the external cert), we should unfortunately still put ips as both `IP:` (for firefox) and `DNS:` entries (for  chrome)

isaac-s: @tehasdf to summarize: everything in `ips` should end up in `dns` as well, but not the other way around, correct?

tehasdf: yeah, that's right

isaac-s:
@tehasdf added another commit plus a test case

tehasdf:
LGTM; tested it by just running a bootstrap with dummy new names/ips provided for the new inputs and checking that they're in the resulting certs

mcouthon:
@isaac-s Can you please unassign me from the PR? :)

isaac-s:
@mcouthon done :-)

PR #644, opened 2017-08-30
CFY-7200 Add a `manager_networks` bootstrap input

mcouthon:
This input will be set in `cloudify.cloudify_agent.networks`,
and subsequently in the provider context.

### DIFF
--- a/inputs/manager-inputs.yaml
+++ b/inputs/manager-inputs.yaml
@@ -30,6 +30,18 @@ inputs:
     type: boolean
     default: true
 
+  #############################
+  # Network configurations
+  #############################
+
+  manager_networks:
+    description: >
+      A list of network names and IP addresses associated with them.
+      By default, there is only a "default" network, with the manager's
+      private IP associated with it. This network can be overwritten
+    default:
+      - default: { get_input: private_ip }
+
   #################################
   # REST Configuration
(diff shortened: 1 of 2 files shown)


types/manager-types.yaml
```diff
@@ -82,6 +82,7 @@ node_types:
             max_workers: 5
             remote_execution_port: 22
             user: centos
+            networks: { get_input: manager_networks }
```
tehasdf: also needs to be added to `simple-manager-blueprint.yaml` because the `cloudify_agent` in there overwrites this, and this one is mostly ignored

inputs/manager-inputs.yaml
```diff
@@ -30,6 +30,18 @@ inputs:
...
+  # Network configurations
+  #############################
+
+  manager_networks:
+    description: >
+      A list of network names and IP addresses associated with them.
+      By default, there is only a "default" network, with the manager's
+      private IP associated with it. This network can be overwritten
+    default:
+      - default: { get_input: private_ip }
```
tehasdf: double `default`? or is this the right syntax?

mcouthon: The double default is because we want the default network to be called “default” as well

PR #649, opened 2017-09-04
CFY-7218: fix nginx validation

isaac-s:


isaac-s:
@mcouthon this PR is against 4.1.1.1-build. Should we merge into it?

mcouthon:
I don't have a problem with that. This doesn't mean we'll release it for them necessarily, but if enough fixes mount up, we might give it a go.

PR #651, opened 2017-09-04
CFY-7203 Clean old upgrade and rollback code

mcouthon:


components/postgresql/scripts/creation_validation.py
```diff
@@ -1,16 +0,0 @@
...
-ctx.download_resource(
-    join('components', 'utils.py'),
-    join(dirname(__file__), 'utils.py'))
-import utils  # NOQA
-
-
-PS_SERVICE_NAME = 'postgresql-9.5'
-
-if utils.is_upgrade:
-    utils.systemd.verify_alive(PS_SERVICE_NAME, append_prefix=False)
```
tehasdf: this is curious actually, i don't think upgrade ever worked in 4.0+, where nginx even existed? :P

components/rabbitmq/scripts/creation_validation.py
```diff
@@ -1,25 +0,0 @@
...
-from cloudify import ctx
-
-ctx.download_resource(
-    join('components', 'utils.py'),
-    join(dirname(__file__), 'utils.py'))
-import utils  # NOQA
-
-runtime_props = ctx.instance.runtime_properties
-
-IMMUTABLE_PROPERTIES = [
```
tehasdf: we had a good idea with this, check that rabbitmq creds didn't change, because then agents won't be able to connect, for an in-place upgrade

components/utils.py
```diff
@@ -1175,104 +1121,13 @@ def get(self, service_name):
...
-    # This function should only be used when during upgrade workflow execution
-    def load_rollback_props(self, service_name):
-        upgrade_props_file = self._get_rollback_props_file_path(service_name)
-        if os.path.isfile(upgrade_props_file):
-            with open(upgrade_props_file) as f:
-                return json.load(f)
-        else:
-            ctx.logger.debug('Failed loading rollback properties. '
-                             'Properties file does not exist.')
+        return self._properties[service_name]
```
tehasdf: remember that `.create()` and `.get()` are likely to be in different scripts, you need to still go through files, the instance won't just be sitting in memory for the whole duration of the bootstrap

components/utils.py
```diff
@@ -1587,143 +1332,22 @@ def wait_for_workflow(deployment_id,
...
-                .format(res.readlines())
-            ctx.logger.error(err)
-            sleep(2)
-        else:
-            return res
-    ctx.abort_operation(err)
-
-
-def create_maintenance_headers(upgrade_props=True):
+def create_maintenance_headers():
```
tehasdf: is this actually still needed? this is only used for waiting for sanity execution status

components/utils.py
mcouthon: Oh, that's right. So I'd need to use a file

components/rabbitmq/scripts/creation_validation.py
mcouthon: Perhaps, but this was never used, right?

components/postgresql/scripts/creation_validation.py
mcouthon: Was never used :)

components/utils.py
mcouthon: Wil check

components/rabbitmq/scripts/creation_validation.py
tehasdf: well it was back in 3.4..; we'll probably have to do something like this again when we do the new "managed" in-place upgrade (if..)

mcouthon: Yeah, maybe, but keeping legacy code for reference seems like a bad idea

tehasdf: oh, i'm not suggesting we keep it, just commenting on it ;P

PR #652, opened 2017-09-05
CFY-7190 Use a separate CA cert

tehasdf:
This also handles CFY-7191 (sign the new certs using the CA certs).

Instead of creating just an internal cert, and using that for everything,
we now separately create a CA cert, which is used (by the manager,
and by the agents) to verify connections, and an internal cert, which
is used by the servers.

This also contains several seemingly unrelated changes that were
required to do this sanely (but increase the overall quality IMO!),
so ideally go commit-by-commit while reviewing this.

### DIFF
--- a/components/manager/scripts/configure_manager.py
+++ b/components/manager/scripts/configure_manager.py
@@ -82,6 +82,6 @@ def init_cloudify_user():
     create_sudoers_file_and_disable_sudo_requiretty()
 
 
+init_cloudify_user()
 configure_security_properties()
 create_certs()
-init_cloudify_user()
--- a/components/manager/scripts/configure_manager.py
+++ b/components/manager/scripts/configure_manager.py
@@ -55,6 +55,7 @@ def create_certs():
     utils.generate_ca_cert()
     networks = ctx_properties['cloudify']['cloudify_agent']['networks']
     internal_rest_host = ctx.instance.runtime_properties['internal_rest_host']
+    utils.store_cert_metadata(internal_rest_host, networks)
     cert_ips = [internal_rest_host] + list(networks.values())
     utils.generate_internal_ssl_cert(ips=cert_ips, name=internal_rest_host)
 
--- a/components/manager/scripts/configure_manager.py
+++ b/components/manager/scripts/configure_manager.py
@@ -52,4 +52,14 @@ def configure_security_properties():
     runtime_props['external_rest_protocol'] = external_rest_protocol
 
 
+def create_certs():
+    utils.mkdir(utils.SSL_CERTS_TARGET_DIR)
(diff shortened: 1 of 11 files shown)


components/utils.py
```diff
@@ -322,58 +338,70 @@ def _generate_ssl_certificate(ips,
...
 distinguished_name = req_distinguished_name
-x509_extensions=SAN
-[ req_distinguished_name ]
-commonName={cn}
-[SAN]
+req_extensions = server_req_extensions
+[ server_req_extensions ]
 subjectAltName={metadata}
+[ req_distinguished_name ]
+commonName = _common_name # ignored, _default is used instead
```
tehasdf: this is unfortunately required

components/utils.py
```diff
@@ -322,58 +338,70 @@ def _generate_ssl_certificate(ips,
...
+    # Remove duplicates from ips
+    cert_metadata = _format_ips(ips)
     ctx.logger.debug('Using certificate metadata: {0}'.format(cert_metadata))
 
-    cert_path = os.path.join(SSL_CERTS_TARGET_DIR, cert_filename)
-    key_path = os.path.join(SSL_CERTS_TARGET_DIR, key_filename)
+    csr_path = '{0}.csr'.format(cert_path)
 
     with tempfile.NamedTemporaryFile(delete=False) as conf_file:
         conf_file.write("""
```
mcouthon: Maybe we can move this string outside of the function?

components/manager/scripts/configure_manager.py
```diff
@@ -56,12 +56,22 @@ def _configure_security_properties():
     runtime_props['external_rest_protocol'] = external_rest_protocol
 
 
+def _create_certs():
+    utils.mkdir(utils.SSL_CERTS_TARGET_DIR)
+    utils.generate_ca_cert()
+    networks = ctx_properties['cloudify']['cloudify_agent']['networks']
+    internal_rest_host = ctx.instance.runtime_properties['internal_rest_host']
+    cert_ips = [internal_rest_host] + list(networks.values())
```
mcouthon: This will be deduplicated, right?

components/manager/scripts/configure_manager.py
```diff
@@ -63,11 +63,34 @@ def _create_certs():
     utils.generate_internal_ssl_cert(ips=cert_ips, name=internal_rest_host)
 
 
+def create_cloudify_user():
```
mcouthon: Go ahead and close this one https://cloudifysource.atlassian.net/browse/CFY-7209 :)

components/manager/scripts/configure_manager.py
tehasdf: yeah

PR #658, opened 2017-09-11
CFY-7188 Add a script to set manager SSL state

tehasdf:
Instead of calling `sed` from the restservice code, let's add
a script that will do the necessary config changes.
This way, we can also run that script from within cluster code,
and also we won't need to enable `sudo` for `sed`, only for this
specific script.

### DIFF
--- a/components/restservice/scripts/set-manager-ssl.py
+++ b/components/restservice/scripts/set-manager-ssl.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python
+
+import argparse
+import subprocess
+
+
+DEFAULT_CONF_PATH = '/etc/nginx/conf.d/default.conf'
+HTTP_PATH = '/etc/nginx/conf.d/http-external-rest-server.cloudify'
+HTTPS_PATH = '/etc/nginx/conf.d/https-external-rest-server.cloudify'
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--ssl-enabled', dest='ssl', action='store_true')
+parser.add_argument('--ssl-disabled', dest='ssl', action='store_false')
+parser.set_defaults(ssl=True)
+
+
+def set_nginx_ssl(enabled):
+    with open(DEFAULT_CONF_PATH) as f:
+        config = f.read()
+    if enabled:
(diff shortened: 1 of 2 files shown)


components/restservice/scripts/set-manager-ssl.py
```diff
@@ -0,0 +1,42 @@
...
+    with open(DEFAULT_CONF_PATH, 'w') as f:
+        f.write(config)
+
+
+def restart_nginx():
+    # wait one second before restarting nginx so that the caller has a chance
+    # to clean up (this is most likely to be called from the REST service)
+    subprocess.check_call([
+        '/usr/bin/systemd-run', '--on-active=1s',
+        '--timer-property=AccuracySec=100ms',
```
tehasdf: the default accuracy is 1 minute, which means `--on-active=1s` would run in anywhere between 1 and 61 seconds...

components/restservice/scripts/set-manager-ssl.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
+
+import argparse
+import subprocess
+
+
+DEFAULT_CONF_PATH = '/etc/nginx/conf.d/default.conf'
+HTTP_PATH = '/etc/nginx/conf.d/http-external-rest-server.cloudify'
+HTTPS_PATH = '/etc/nginx/conf.d/https-external-rest-server.cloudify'
```
yarivgiga: if you have the script on cloudify-manager-blueprints repo, then you don't have to literally write the consts here.
I think it would be better to render them while deploying the script.
Checkout the preconfigure.py of nginx.

tehasdf: this is a good idea, but looks like overkill right now: if we wanted to render the script currently, we'd have to add the paths to runtime_properties, and it feels wrong to add things there to just render paths...

also, other places (like eg. nginx's default.conf, and well preconfigure.py as well) also have those paths just as strings hardcoded in them; so I think, if we want to make those paths configurable, let's make it as a separate issue, addressing all of these places (and this one as well)

PR #660, opened 2017-09-16
CFY-6967: removed unnecessary moving of logrotate

isaac-s:
1. Nobody seems to know why this block exists.
2. It causes bootstrap to fail if permissions of `/etc/cron.hourly` is `700`.


tehasdf (approved):
Checking this hourly is very much overkill, with the `size` setting (100M) that we use. Daily is enough, so indeed we can simply skip this.

mcouthon (approved):

PR #666, opened 2017-09-18
CFY-7253: remove redirection

isaac-s:


isaac-s:
We disqualified 444 because it's nginx-specific.

PR #668, opened 2017-09-22
CFY-5808: Replaced CLI install functionality

isaac-s:
Current functionality is completely broken.


isaac-s:
@mcouthon I agree, when do you think we'll have the CLI included in the single tar? I can work on all the rest.

mcouthon:
@isaac-s I will push for it immediately. When will this actually happen I can't say. Possibly not until 4.3.

isaac-s:
Someone else has already pushed a change.

PR #686, opened 2017-10-10
Cfy 7326 delete logs and events periodically

adigrabow:


components/restservice/scripts/configure.py
```diff
@@ -185,13 +185,24 @@ def _allow_creating_cluster():
     utils.allow_user_to_sudo_command(command, description)
 
 
+def _deploy_db_cleanup_script():
+    """ Copy the script that deletes logs and events from the
+    Cloudify DB to /etc/cloudify, so it will be available to the users."""
+
+    source = '/opt/cfy/cloudify-manager-blueprints/components/restservice/' \
```
dusking: should be: join(CONFIG_PATH, 'delete_logs_and_events_from_db.py')

components/restservice/scripts/configure.py
```diff
@@ -185,13 +185,24 @@ def _allow_creating_cluster():
...
+             'scripts/delete_logs_and_events_from_db.py'
+    destination = 'etc/cloudify'
```
dusking: **/etc**/cloudify/

PR #689, opened 2017-10-15
CFY-7362 validate openssl

dusking:


components/manager/scripts/validate.py
```diff
@@ -131,13 +132,32 @@ def _validate_resources_package_url(manager_resources_package_url):
...
+        import ssl
+        version = ssl.OPENSSL_VERSION.split()[1]
+        if LooseVersion(version) < LooseVersion(required_version):
+            return _error(
+                "Cloudify Manager requires OpenSSL {0}, current version: {1}"
+                "".format(required_version, version))
+    except ImportError:
+        return _error(
+            "Cloudify Manager requires OpenSSL {0}".format(required_version))
+    except Exception as ex:
```
mcouthon: Is this necessary? What other exception can there be?

PR #691, opened 2017-10-15
CFY-7296 Add set_global permissions to authorization.conf

inbala:


components/restservice/config/authorization.conf
```diff
@@ -67,6 +67,8 @@ permissions:
     - sys_admin
     - manager
     - user
+  blueprint_set_global:
```
dusking: maybe resource_set_global is sufficient?

PR #693, opened 2017-10-16
CFY-7236 Allow passing internal certs in inputs

tehasdf:
This adds the inputs `ca_certificate`, `ca_key`, `internal_certificate` and `internal_key`.

The user can now pass the internal CA cert+key, and the internal cert+key.

Optionally, the user can choose to not provide the CA key, but only the cert.
In this case, it is required that the user also provides the internal cert+key,
because they can't be generated without a CA key.

This means it is now possible to generate all the required certificates before
bootstrapping a manager, and cloudify does not require the CA key that was
used to generate the keys.

### DIFF
--- a/components/nginx/scripts/preconfigure.py
+++ b/components/nginx/scripts/preconfigure.py
@@ -83,6 +83,8 @@ def _deploy_nginx_config_files():
 
 def _deploy_cert_and_key(cert_src, key_src, cert_path, key_path):
     def _try_deploy(src, dest):
+        if not src:
+            return False
         try:
             utils.deploy_blueprint_resource(
                 src_runtime_props[src], dest, NGINX_SERVICE_NAME,
--- a/components/nginx/scripts/preconfigure.py
+++ b/components/nginx/scripts/preconfigure.py
@@ -80,21 +80,28 @@ def _deploy_nginx_config_files():
         )
 
 
+def _deploy_cert_and_key(cert_src, key_src, cert_path, key_path):
+    def _try_deploy(src, dest):
+        try:
+            utils.deploy_blueprint_resource(
+                src_runtime_props[src], dest, NGINX_SERVICE_NAME,
+                user_resource=True, load_ctx=False)
+            return True
+        except Exception as e:
+            if "No such file or directory" in e.stderr:
+                return False
+            else:
+                raise
+    return _try_deploy(cert_src, cert_path), _try_deploy(key_src, key_path)
+
+
 def _deploy_external_cert():
     target_runtime_props = ctx.target.instance.runtime_properties
 
-    external_key_deployed = utils.deploy_blueprint_resource(
-        src_runtime_props['rest_key'],
-        utils.EXTERNAL_KEY_PATH,
-        NGINX_SERVICE_NAME,
-        user_resource=True, load_ctx=False
-    )
-    external_cert_deployed = utils.deploy_blueprint_resource(
-        src_runtime_props['rest_certificate'],
-        utils.EXTERNAL_CERT_PATH,
-        NGINX_SERVICE_NAME,
-        user_resource=True, load_ctx=False
-    )
+    external_cert_deployed, external_key_deployed = _deploy_cert_and_key(
+        'rest_certificate', 'rest_key',
+        utils.EXTERNAL_CERT_PATH, utils.EXTERNAL_KEY_PATH)
+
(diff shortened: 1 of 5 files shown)


tehasdf:
Please don't merge yet

components/nginx/scripts/preconfigure.py
```diff
@@ -80,11 +80,54 @@ def _deploy_nginx_config_files():
...
 
+    external_cert_deployed, external_key_deployed = _deploy_cert_and_key(
+        'rest_certificate', 'rest_key',
+        utils.EXTERNAL_CERT_PATH, utils.EXTERNAL_KEY_PATH)
+
+    if external_key_deployed and external_cert_deployed:
+        ctx.logger.info(
+            'Deployed user-provided external SSL certificate and private key')
+    elif not external_cert_deployed and not external_key_deployed:
+        utils.generate_ssl_certificate(
```
mcouthon: Let's maybe take it out to its own function. It's getting a little crowded here.

components/nginx/scripts/preconfigure.py
```diff
@@ -80,11 +80,54 @@ def _deploy_nginx_config_files():
...
+            utils.deploy_blueprint_resource(
+                src_runtime_props[src], dest, NGINX_SERVICE_NAME,
+                user_resource=True, load_ctx=False)
+            return True
+        except Exception as e:
+            if "No such file or directory" in e.stderr:
+                return False
+            else:
+                raise
+    return _try_deploy(cert_src, cert_path), _try_deploy(key_src, key_path)
```
mcouthon: Very clever 😛

components/nginx/scripts/preconfigure.py
```diff
@@ -80,11 +80,54 @@ def _deploy_nginx_config_files():
...
+                            'key must be provided, or neither. Only the {0} '
+                            'was provided'.format(what_deployed))
+
+    src_runtime_props['external_cert_path'] = utils.EXTERNAL_CERT_PATH
+    src_runtime_props['external_key_path'] = utils.EXTERNAL_KEY_PATH
+
+
+def preconfigure_nginx():
     # This is used by nginx's default.conf to select the relevant configuration
+    target_runtime_props = ctx.target.instance.runtime_properties
```
mcouthon: No reason to have 2 of those. Let's just have one on the module level.

components/nginx/scripts/preconfigure.py
```diff
@@ -118,14 +151,34 @@ def preconfigure_nginx():
 
 def create_certs():
     utils.mkdir(utils.SSL_CERTS_TARGET_DIR)
-    utils.generate_ca_cert()
+    ca_cert_deployed, ca_key_deployed = _deploy_cert_and_key(
+        'ca_certificate', 'ca_key',
+        utils.INTERNAL_CA_CERT_PATH, utils.INTERNAL_CA_KEY_PATH)
+    has_ca_key = ca_key_deployed
+    if not ca_cert_deployed:
```
mcouthon: This is getting a bit convoluted. Let's add some comments here. I know the code should be readable enough by itself, but I'm afraid future generations will have trouble with it.

components/nginx/scripts/preconfigure.py
```diff
@@ -118,14 +151,34 @@ def preconfigure_nginx():
...
         ctx.target.instance.runtime_properties['internal_rest_host']
     utils.store_cert_metadata(internal_rest_host, networks)
-    cert_ips = [internal_rest_host] + list(networks.values())
-    utils.generate_internal_ssl_cert(ips=cert_ips, name=internal_rest_host)
+
+    internal_cert_deployed, internal_key_deployed = _deploy_cert_and_key(
+        'internal_certificate', 'internal_key',
+        utils.INTERNAL_CERT_PATH, utils.INTERNAL_KEY_PATH)
+
+    if not internal_cert_deployed and not internal_key_deployed:
```
mcouthon: Ditto re comments.

components/nginx/scripts/preconfigure.py
tehasdf: added a few comments (didn't affect this very line so github won't hide this message)

components/nginx/scripts/preconfigure.py
tehasdf: added a few comments (didn't affect this very line so github won't hide this message)

PR #694, opened 2017-10-17
CFY-7380 fix openssl version validation

dusking:


components/manager/scripts/validate.py
```diff
@@ -135,8 +135,10 @@ def _validate_resources_package_url(manager_resources_package_url):
 def _validate_openssl_version(required_version):
     ctx.logger.info('Validating OpenSSL version...')
     try:
-        import ssl
-        version = ssl.OPENSSL_VERSION.split()[1]
+        # we can't just `import ssl` and check the version
+        # because sometimes python is referencing to the old ssl version
+        output = os.popen('openssl version').read()
```
mcouthon: Shouldn't we use `utils.run`?

dusking: tried it, too much hassle to add it to the tests framework..

PR #706, opened 2017-10-25
CFY-7423: validate pre-existence of certain packages

isaac-s:


components/python/scripts/bootstrap_validate.py
```diff
@@ -0,0 +1,41 @@
...
+            'python-backports-ssl_match_hostname']:
+    ctx.logger.info('Ensuring {0} is installed'.format(pkg))
+    is_installed = utils.RpmPackageHandler.is_package_installed(pkg)
+    if not is_installed:
+        missing_packages.add(pkg)
+
+if missing_packages:
+    ctx.abort_operation('Prerequisite packages missing: {0}. '
+                        'Please ensure these packages are installed and '
+                        'try again'.format(list(missing_packages)))
```
tehasdf: you can use `', '.join(missing_packages)` instead of `list(missing_packages)` to have it formatted nicer - commaseparated instead of python's repr

tehasdf:
travis flake8 failure is not related to this change, but is only surfaced because of a recent flake8 version update

will be addressed as a separate PR

isaac-s:
Merged as part of https://github.com/cloudify-cosmo/cloudify-manager-blueprints/pull/714

PR #707, opened 2017-10-25
Fix bare except flake8 complaint

geokala:


components/utils.py
```diff
@@ -562,7 +562,7 @@ def download_file(url, destination=''):
 def get_file_name_from_url(url):
     try:
         return url.split('/')[-1]
-    except:
+    except Exception:
```
tehasdf: i don't think this is required at all, because i don't think `url.split('/')[-1]` can ever throw anyway

components/utils.py
```diff
@@ -552,7 +552,7 @@ def download_file(url, destination=''):
             f = urllib.URLopener()
             # TODO: try except with @retry
             f.retrieve(final_url, destination)
-        except:
+        except Exception:
```
tehasdf: this should probably use some other exception (possibly IOError)

components/utils.py
funkyHat: (Also we could just use `os.path.basename` here...)

PR #709, opened 2017-10-26
CFY-6300: add ssh user validation

isaac-s:


isaac-s:
Committed as part of https://github.com/cloudify-cosmo/cloudify-manager-blueprints/pull/714

PR #716, opened 2017-11-01
CFY-7437 Use API token for stage auth

geokala:


components/stage/scripts/create.py
```diff
@@ -93,6 +101,18 @@ def _install_stage():
...
+def configure_script(script_name, description):
+    utils.deploy_sudo_command_script(
+        script_name,
+        description,
+        component=SERVICE_NAME,
+        allow_as=STAGE_USER,
+    )
+
+    utils.chmod('a+rx', '/opt/cloudify/stage/' + script_name)
+    utils.sudo(['usermod', '-aG', utils.CLOUDIFY_GROUP, STAGE_USER])
```
tehasdf: this usermod line should only be called once, not inside the function

components/stage/scripts/make-auth-token.py
```diff
@@ -0,0 +1,52 @@
+from manager_rest import config, server, storage
+
+from contextlib import contextmanager
+import json
+import logging
+import subprocess
+
+RESTSERVICE_CONFIG_PATH = '/opt/manager/cloudify-rest.conf'
+MANAGER_PYTHON = '/opt/manager/env/bin/python'
+GETENCUSER_SCRIPT = '/opt/mgmtworker/env/lib/python2.7/site-packages/cloudify_system_workflows/snapshots/getencodeduser.py'  # noqa
```
tehasdf: the `getencodeuser.py` script is a hack to allow mgmtworker (workflow) to call a function from manager_rest

so using that script here means we're using a hack to call a hack.. ;P

we're already using manager_rest here, so maybe let's use the idencoder function directly? ie. copy the call directly into this script:
```python
from manager_rest.storage import idencoder
encoded_uid = idencoder.get_encoder().encode(user_id)
```
(probably will need similar config dance as the storage manager

tehasdf: if we end up using that, we'll likely want to run the whole script under a `with app_context()` so that both idencoder and storagemanager are there

probably will be nicer than fine-grained withs

components/stage/scripts/make-auth-token.py
```diff
@@ -0,0 +1,41 @@
+from manager_rest import config, server, storage
+
+import os
+
+RESTSERVICE_CONFIG_PATH = '/opt/manager/cloudify-rest.conf'
+RESTSEC_CONFIG_PATH = '/opt/manager/rest-security.conf'
+MANAGER_PYTHON = '/opt/manager/env/bin/python'
+GETENCUSER_SCRIPT = '/opt/mgmtworker/env/lib/python2.7/site-packages/cloudify_system_workflows/snapshots/getencodeduser.py'  # noqa
```
tehasdf: unused - remove

components/stage/scripts/create.py
```diff
@@ -93,6 +107,17 @@ def _install_stage():
...
 
+def configure_script(script_name, description):
+    utils.deploy_sudo_command_script(
+        script_name,
+        description,
+        component=SERVICE_NAME,
+        allow_as=STAGE_USER,
+    )
+    utils.chmod('a+rx', '/opt/cloudify/stage/' + script_name)
+    utils.sudo(['usermod', '-aG', utils.CLOUDIFY_GROUP, STAGE_USER])
```
tehasdf: sure we want to call this for every script?

geokala: We do for every script that currently uses this, yes.

PR #724, opened 2017-11-20
Cfy 7577 properly restore stage files

geokala:


components/stage/scripts/restore-snapshot.py
```diff
@@ -4,22 +4,26 @@
 import shutil
 import argparse
 
-HOME_DIR = "{{ ctx.instance.runtime_properties.home_dir}}"
+HOME_DIR = "/opt/cloudify-stage"
```
dusking: {{ ctx.instance.runtime_properties.home_dir}} is /opt/cloudify-stage...

PR #725, opened 2017-11-21
CFY-7554 Add roles for setting the availability of a resource

inbala:


components/restservice/config/authorization.conf
```diff
@@ -459,6 +466,17 @@ permissions:
...
+  resource_set_availability:
+    - sys_admin
+    - manager
+    - user
+  deployment_set_availability:
+    - sys_admin
+    - manager
+    - user
+    - operations
+  snapshot_set_availability:
```
dusking: no need to set availability to snapshots.. it's relevant only to sys_admin..

PR #727, opened 2017-12-08
CFY-7645: add https prefixes

isaac-s:
@mcouthon you should probably update the new RPM-based installer accordingly...


tehasdf:
cloudify-cosmo/cloudify-manager-install#53 is the corresponding cloudify-manager-install PR

